Issue #4, opened 2019-08-06
optional properties?


danbrauer:
Hi!  It is possible to define an optional object property?

What I mean is something that, in Typescript, would be defined like `b` below:

```
type MyType = {
   a: string,
   b?: string 
}
```

In this case, `a` must be there but `b` may be absent.

We at first thought this could be defined with something like this, with typebox:

```
const MyTypeSchema = Type.Object({
   a: Type.String(),
   b: Type.Optional(Type.String())
})
```

But that's not quite the same.  It leads to a type like this:

```
type MyType = {
   a: string,
   b: string  // <--- we'd want `b?: string`
}
```

Just wondering if we're missing something with respect to optional properties, or if the project does not allow them?

Thank you!
Dan

sinclairzx81:
Hi,

Yeah, good spotting :) Its actually something that needs doing. Both Optional and ReadOnly modifiers need implementation, with Optional being particularly relevant to JSONSchema.

Had a couple of attempts at puzzling through it, and recall there might be a need for an additional support type outside of `TBase<T>` to hang `TOptional<TBase<T>>` off. As of this version of, Optional maps to `T | undefined`.

You're welcome to have a try solving it. Any PR would be hugely welcome.

sinclairzx81:
https://github.com/microsoft/TypeScript/issues/24951

danbrauer:
I won't be able to try a PR anytime soon, but thank you for replying and confirming.  --Dan

webberwang:
The issue was closed with any resolution? doesn't seem possible to do

```typescript
interface User {
  name?: string
}
```

We can only do

```typescript
const User = Type.Object({
  name: Type.Optional(Type.String())
})

interface User {
  name?: string | undefined
}
```

sinclairzx81:
@webberwang Hi, This is a very old issue :) 

You can actually configure TypeBox to assert for the following
```typescript
interface User {
  name?: string
}
```
This is configured via the `TypeSystemPolicy.ExactOptionalPropertyTypes` setting.
```typescript
import { TypeSystemPolicy } from '@sinclair/typebox/system'
import { Value } from '@sinclair/typebox/value'
import { Type } from '@sinclair/typebox'

TypeSystemPolicy.ExactOptionalPropertyTypes = true

const T = Type.Object({
   x: Type.Optional(Type.String())
})

Value.Check(T, { x: 'hello' })    // true
Value.Check(T, { })               // true
[...]

PR #6, opened 2020-02-29
add DateTime type

jhewlett:
If you're open to it, this adds support for the `date-time` format (https://json-schema.org/understanding-json-schema/reference/string.html#dates-and-times)

I'm open to other designs.

I also explored adding an optional `format` arg to `String`, but then realized there were already dedicated types for things like `Pattern` and `Range`.

I also explored adding something like `StringFormat` that would take in `date-time` for now and could later support `date`, `time`, or any of the other formats. That might scale better if we want to add more formats in the future.


sinclairzx81:
Hi, Thanks ! This is a very clean PR :)

Historically, Patterns like DateTime (ISO8601) and other string formats were implemented outside of TypeBox in the following way.

```typescript
// RFC4122
const Guid = Type.Pattern(/^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/)

// RFC3339 | ISO8601
const DateTime = Type.Pattern(/(\d{4})-(\d{2})-(\d{2})T(\d{2})\:(\d{2})\:(\d{2})[+-](\d{2})\:(\d{2})/)

const Record = Type.Object({
    id: Guid,
    created: DateTime
})
```
TypeBox was written to try and align with the 'lowest common denominator' with respect to validation. Unlike similar libraries (`io-ts` and `runtypes`) TypeBox relies heavily on leveraging external JSON Schema validation rather than bringing its own validator. As such TypeBox is a bit more concerned with the down level capabilities of the JSON validator. In the wild, TypeBox is used to provide HTTP RPC reflection of interfaces where the consumers are .NET, Java and Python processes. Their validation capabilities may not align with draft 6 | 7 functionality. 

However, given how frequent `DateTime` is passed over the wire, I do actually like having DateTime expressed out of the box (especially as its represented in more modern drafts) but perhaps expressed as a `Pattern` via Regular Expressions to allow lesser validators to make use of it. 

So,
```typescript
public DateTime(): TString {
   return { type: "string", format: "date-time" } as any as TString
}
```
Would become
```typescript
public DateTime(): TString {
    return this.Pattern(/(\d{4})-(\d{2})-(\d{2})T(\d{2})\:(\d{2})\:(\d{2})[+-](\d{2})\:(\d{2})/)
}
```

Looking at https://json-schema.org/understanding-json-schema/reference/string.html#dates-and-times, It does seem as tho `date`, `time`, `email`, `idn-email`, `hostname`, `idn-hostname`, `uri` etc are all available in draft 7. It is very tempting to implement all of these internal to TypeBox as Regex patterns. TypeBox was written prior to draft 7, it would be good to try and bring it inline with the newer draft versions in terms of functionality while still accommodating pre draft-7 validators.

What are your thoughts on moving `DateTime` to a Pattern as above? I am certainly open to discussing this further. My thinking is that TypeBox could express these modern formats through Regex today, and through `{format: 'x'}` approach tomorrow once the IETF ratifies JSON schema.

Look forward to getting your thoughts

S

jhewlett:
Hi,

Thanks for the background, I'm not too familiar with the different stages of the spec.

My concern with using regex is that it may not be exactly the same behavior as you'd get with `format`. For example, the regex you posted for ISO8601 doesn't allow a `Z` for the timezone. It may be close enough for most uses, but then I'd argue it's better left for the user to implement themselves with a `Pattern` rather than being built-in.

So I guess I think it would be cool if there was a good, clear way to "opt-in" to using newer features with the understanding that they may not be implemented in all validators.

If you don't want to do that, then I kind of think it's better left for users to implement. Perhaps we could add some documentation to show examples of how you can approximate `date-time` with a `Pattern`.

fox1t:
Hi. I just discovered this repository (kudos to @jhewlett :) )and I really like what it does. 
Regarding this topic, I think having integrated a way to check common types would be nice, since it could lower the possibilities on making wrong regexp by final users. What I find awesome about this lib, as @jhewlett pointed out [here](https://github.com/fastify/fastify/issues/1668#issuecomment-593548468), is that, thanks to its compliance to both JSONSchema and TS types, it plays really well with Fastify .

What do you think if we add some common regexp as "built-in patterns", so users can import them directly?

```typescript
import {GuidPattern, DateTimePattern} from 'typebox'

const Record = Type.Object({
    id: GuidPattern,
    created: DateTimePattern
})
```
We should add more specific `formats` as opt-in is someone wants to use drafts > 6 afterwards.

sinclairzx81:
Hi all, Apologies for the delay. Have been spending some time thinking about this and have a possible suggestion.

```typescript
import { Type, Static } from 'typebox'

const Record = Type.Object({
     created: Type.Format('date-time'), // TString  | { "format": "date-time" }
     website: Type.Format('url'),       // TString  | { "format": "url" }
     email:   Type.Format('email')      // TString  | { "format": "email" }
})
```
This would allow users to opt in to draft 6, draft 7 and future `format` functionality (assuming the `format` remains a plain identifier). This somewhat aligns with the current `Type.Pattern(....)` and future `Type.Regex(....)`

Thoughts?

fox1t:
> Hi all, Apologies for the delay. Have been spending some time thinking about this and have a possible suggestion.
> 
> ```ts
> import { Type, Static } from 'typebox'
> 
> const Record = Type.Object({
>      created: Type.Format('date-time'), // TString  | { "format": "date-time" }
>      website: Type.Format('url'),       // TString  | { "format": "url" }
>      email:   Type.Format('email')      // TString  | { "format": "email" }
> })
> ```
> 
> This would allow users to opt in to draft 6, draft 7 and future `format` functionality (assuming the `format` remains a plain identifier). This somewhat aligns with the current `Type.Pattern(....)` and future `Type.Regex(....)`
> 
> Thoughts?

No worries about the delay (if we can call like that). I really like your proposal!

jhewlett:
@sinclairzx81 I like that a lot.

Curious, what's the reason for having the type be `TString  | { "format": "date-time" }`? Just as a way to distinguish it from other `Format` instances?

If you look at, say, `Range`, or `Pattern`, they're just `TString`.

sinclairzx81:
@jhewlett @fox1t 

I have pushed a new branch [`typebox-next`](https://github.com/sinclairzx81/typebox/tree/typebox-next) where I have implemented the `Type.Format('date-time')` interface as well as re-implemented the TypeBox type mappings from the ground up. I have also added a few other features such as `Type.Optional(...)` and `Type.Readonly(...)` which were in the list of things that needed doing for this project (https://github.com/sinclairzx81/typebox/issues/4). (I haven't really looked at this project in over a year, so it needed doing)

Note, the [`typebox-next`](https://github.com/sinclairzx81/typebox/tree/typebox-next) likely needs an up to date version of the TypeScript compiler to do its thing. So will leave it in the `next` branch for the next few days while I integrate it with some other software I'm working on, but If you want to try it out your side you're more than welcome. (would appreciate any feedback actually).

@jhewlett

Still happy to accept this PR on the master branch with the small change from `Type.DateTime()` to `Type.Format('date-time')` as mentioned above. This project is in some need of contributors and do appreciate the PR very much. Let me know if you want to push on through with the PR.

Many Thanks
S

jhewlett:
@sinclairzx81 Sure, I'll make that change and push up another commit.

sinclairzx81:
Thanks, Looks good to me :)

PR #14, opened 2020-03-25
Allow format keyword to be augmented with custom formats

matomesc:
Allows consumers to augment the `format` keyword, for example:

```ts
declare module '@sinclair/typebox' {
  interface UserDefinedOptions {
    format: 'custom-format-1' | 'custom-format-2';
  }
}
```

### DIFF
--- a/src/typebox.ts
+++ b/src/typebox.ts
@@ -179,8 +179,8 @@ export type StringOptions = {
   minLength?: number
   maxLength?: number
   pattern?: string
-  format?: FormatOption
-} & UserDefinedOptions
+  format?: FormatOption | UserDefinedOptions['format']
+} & Omit<UserDefinedOptions, 'format'>
 
 export type TLiteral = TStringLiteral<string> | TNumberLiteral<number> | TBooleanLiteral<boolean>
 export type TStringLiteral<T> = { type: 'string', enum: [T] } & UserDefinedOptions
--- a/src/typebox.ts
+++ b/src/typebox.ts
@@ -175,12 +175,17 @@ export type NumberOptions = {
   multipleOf?: number
 } & UserDefinedOptions
 
-export type StringOptions = {
-  minLength?: number
-  maxLength?: number
-  pattern?: string
-  format?: FormatOption | UserDefinedOptions['format']
-} & Omit<UserDefinedOptions, 'format'>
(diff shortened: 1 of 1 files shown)


sinclairzx81:
Hi, thanks for the PR! :)

Just had a look through and note that the changes do make the `format` property an `any?` type instead of the `FormatOption` whose string union is derived from [here](https://json-schema.org/understanding-json-schema/reference/string.html#dates-and-times). I'm fairly keen to retain those string options if possible, or at a minimum keep the format option narrowed to type `string` instead of `any`.

screenshot


I have mulled over this a few times when putting typebox together and tried a few approaches. Unfortunately, the code below loses information about `x, y and z`, so the current implementation of typebox just defaults to the documented format options and assumes custom formats would be implemented through `Pattern` regex.

```typescript
type A = 'x' | 'y' | 'z' | string // A is just string :(
```
However, if you have custom format options known to your validator. How would you feel about the following?
```typescript
export type FormatOption =
  | 'date-time' | 'time' | 'date' | 'email' | 'idn-email' | 'hostname'
  | 'idn-hostname' | 'ipv4' | 'ipv6' | 'uri' | 'uri-reference' | 'iri'
  | 'iri-reference' | 'uri-template' | 'json-pointer' | 'relative-json-pointer'
  | 'regex' | 'custom' // <- add custom option added here.
```
which would allow for something like..
```typescript
type X = Type.String({ format: 'foo' as 'custom' })
```
That would retain the existing union options, yet still provide a semi reasonable escape hatch for custom formats.

Look forward to getting your thoughts
Thanks Again!
S

matomesc:
Format option is now working with or without augmentation. I had to check if `UserDefinedOptions['format']` is in fact a union since the `UserDefinedOptions` interface maps all keys to `any` type, so when format wasn't augmented then `any | FormatOptions = any`.

sinclairzx81:
Hey Thanks :) Works great. I'll merge this one and apply a couple of updates my side. Would be keen in future to explore if there are more straightforward ways to augment those unions to allow for additional formats though. The `UnionToIntersect<T>` mapping is a lot of hoops to jump through. Ill do some reading up my side.

Many thanks, super appreciated :) !
S

sinclairzx81:

PR #15, opened 2020-03-30
add enum support

geekflyer:
this adds support for enums.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -305,6 +305,33 @@ The following are object property modifiers. Note that `Type.Optional(...)` will
     </tbody>
 </table>
 
+### Enums
+
+It is possible to define TypeScript enums and use them as part of your TypeBox schema.
+Both number and string-valued enums are supported.
+
+```ts
+enum Color {
+    Red = 'red',
+    Blue = 'blue'
+}
+
+T = Type.Enum(Color); // -> json-schema: `{ enum: ['red','green'] }`
+```
+
+Note that the generated json-schema will only permit the *values* of the enum, not its *keys*.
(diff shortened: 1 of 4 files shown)


sinclairzx81:
Hi!

This is a neat bit of functionality. I do have a couple of updates id like to make, but ill merge as is and implement those updates over the top. This is great work, thank you very much !

sinclairzx81:

geekflyer:
hey @sinclairzx81 I think your refactor of my code broke some of the type inference for the enum types ;-)

try this with the current master:

```
    enum SomeEnum {
      Foo,
      Bar
    }

    const T = Type.Enum(SomeEnum);

    const bla: Static<typeof T> = SomeEnum.Foo; // Type 'SomeEnum.Foo' is not assignable to type 'typeof SomeEnum'.ts(2322)
```

the reason is that you changed the signature of 
`Enum<T extends Record<keyof T, string | number>>(enumObj: T, options?: UserDefinedOptions): TEnum<T[keyof T]>` which infers to `TEnum<SomeEnum>`
to
`Enum<T extends Record<string, string | number>>(item: T, options?: UserDefinedOptions): TEnum<T>`
to `TEnum<typeof Enum>`.

I can't explain it in detail, but when you pass the `enum object` instead of an enum value to a function, TypeScript infers `T` to be not the enum type, but the type of the enum object, which is different.
That's similar to how class inference works in TypeScript.

sinclairzx81:
Yes, sorry :D Just noticed that.

Just fixed that up for you, resolves back to the enum now, not `typeof T`. My bad :)

Updates in `0.9.11`

Thanks again :D

PR #16, opened 2020-03-30
add uuid-format

geekflyer:
this adds the `uuid` format as per https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.7.3.5 .


sinclairzx81:
Thanks again!

Yes, missed this one. Had been working from [here](https://json-schema.org/understanding-json-schema/reference/string.html#format) as a source of truth and had read somewhere there was contention around `uuid`. Have tested however and it works fine with ajv, and if its in the spec it certainly belongs here. I'll be pleased to retire the `Type.Guid()` in subsequent releases :) 

Again, thank you very much !
S

sinclairzx81:

PR #30, opened 2020-06-02
Fix Optional type should add null type in type array

wtospit:
When using Type.Optional currently It will not add null type in type properties for JSON Schema, so when I used this in Fastify it will use zero value for the given type (for example 0 for Type.Optional(Type.Number()))

This PR. will add null type when using Type.Optional

### DIFF
--- a/spec/object.ts
+++ b/spec/object.ts
@@ -33,38 +33,62 @@ describe('Object', () => {
     fail(T, 123)
     fail(T, null)
   })
-  it('Optional',  () => {
-    const T = Type.Object({
-      a: Type.Optional(Type.Number()),
-      b: Type.Optional(Type.String()),
-      c: Type.Optional(Type.Boolean()),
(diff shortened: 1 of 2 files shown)


sinclairzx81:
Hi, thanks for the PR. 

Have you tried?

```typescript
const T = Type.Object({
   x: Type.Union([Type.Null(), Type.String()])  // null | string 
})
```

TypeBox draws a distinction between `null` and `undefined`. The resolution of a `Type.Optional(...)` should resolve to `T | undefined` (not `T | null`). 

In the context of an `Type.Object(...)` with a `Type.Optional(...)` property, it means "allow the absence of that property on the object". (which implicates the `required: [...]` array of the objects schema).

Note also that, TypeBox treats `null` as a kind of value you can pass on a object. It draws parity with both typescript and JSON schema's representation https://json-schema.org/understanding-json-schema/reference/null.html

wtospit:
Thanks, will try it.

Issue #37, opened 2020-10-18
Type.Union() should result in anyOf instead of oneOf


sunaurus:
This code:

```
const FirstSchema = Type.Object({foo: Type.String()})
const SecondSchema = Type.Object({foo: Type.String(), bar: Type.Optional(Type.String())})
const Union = Type.Union([FirstSchema, SecondSchema]);
```

Generates the following schema:

```
    {
      "oneOf": [
        {
          "type": "object",
          "properties": {
            "foo": {
              "type": "string"
            }
          },
          "required": [
            "foo"
          ]
        },
        {
          "type": "object",
          "properties": {
            "foo": {
              "type": "string"
            },
            "bar": {
              "type": "string"
            }
          },
          "required": [
            "foo"
          ]
        }
      ]
    }
```

The issue is that it uses oneOf instead of anyOf - oneOf means that objects may match **exactly one** schema, but in TypeScript, unions work like anyOf, not oneOf. With anyOf, objects must match **at least one** schema.

For example, this is valid TypeScript:

```
type FirstType = {
  foo: string;
}

type SecondType = {
  foo: string;
  bar?: string;
}

type UnionType = FirstType | SecondType;

const test: UnionType = {foo: "test"};
```
However, `{foo: "test"}` will not match the `Union` schema shown above, because:
```
JSON is valid against more than one schema from 'oneOf'.
```

Would you accept a PR that changes oneOf to anyOf for `Type.Union()`?

sinclairzx81:
@sunaurus Hi, This update has been published to npm along with a variety of other updates to better align with TypeScript 4.0. Thanks for the PR on this. Updates in `@sinclair/typebox@0.12.0`.

Cheers
S

PR #39, opened 2020-10-18
Use anyOf instead of oneOf for Union (sinclairzx81#37)

sunaurus:
This PR has the proposed fix for #37 
Also changed the test cases for Union to reflect the new (correct) behavior. Made sure that new tests would not pass with the old logic.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -244,7 +244,7 @@ TypeBox provides a number of functions to generate JSON Schema data types. The f
         <tr>
             <td>Union</td>
             <td><code>const T = Type.Union([Type.Number(), Type.String()])</code></td>
-            <td><code>{ oneOf: [{ type: 'number'}, {type: 'string'}] }</code></td>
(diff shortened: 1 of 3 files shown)


sinclairzx81:
@sunaurus Hi, thanks for this PR. If I recall, `oneOf` was originally used for the `Type.Union`, but I forget the reason for opting for `anyOf`. Happy to roll this update through, but will want to check some of the downstream regressions potentially caused by this change. 

Also, apologies for the slow response on the PR (I am currently maxing out on some other work at the moment and don't have much bandwidth to spend on typebox currently). Happy to merge this through now, but may defer npm publishing for a few weeks to give me some time to test a few edge cases, I hope you understand. There are actually some additional updates im looking to add (specifically adding `Type.Unknown()` and potentially reconciling the `Function` and `Constructor` contract stuff as inferable on `Type.Object`). Would be keen to roll your update through with those changes.

Again, Many thanks for this PR! :)

Issue #34, opened 2020-09-08
Make use of $ref in JSON Schema


Remirror-zz:
````
const Person = Type.Object({
  'name': Type.String(),
  'age': Type.Number()
});
const ChessMatch = Type.Object({
  'playerWhite': Person,
  'playerBlack': Person
});
console.log(JSON.stringify(ChessMatch));
````
outputs
````
{
  "type": "object",
  "properties": {
    "playerWhite": {
      "type": "object",
      "properties": {
        "name": {"type": "string"},
        "age": {"type": "number"}
      },
      "required": ["name","age"]
    },
    "playerBlack": {
      "type": "object",
      "properties": // same as above
// ...
````

This means, the definition of `Person` is repeated in the output. Instead, please consider using the `$ref` keyword. The output would then look something like this:

````
{
  "type": "object",
  "definitions": {
    "person": {
      "type": "object",
      "properties": {
        "name": {"type": "string"},
        "age": {"type": "number"}
      }
    }
  },
  "properties": {
    "playerWhite": {"$ref": "#/definitions/person"},
    "playerBlack": {"$ref": "#/definitions/person"}
  }
}
````

**Comment by sinclairzx81** [comment]
@Remirror-zz Hi

I like the idea, however there isn't any way to trivially resolve the `"definitions"`.`"person"` property or the `{"$ref": "#/definitions/person"}` as the name "person" isn't directly resolvable from `const Person = Type.Object({ ... })`. TypeBox would need to be extended to support the feature. 

One option may be to extend TypeBox to support something like...

```typescript
const Person = Type.Object({
  'name': Type.String(),
  'age': Type.Number()
}, { ref: 'person' }); // new

const ChessMatch = Type.Object({
  'playerWhite': Person, // check here if Person has a 'ref', and if so assign it to "definitions"
  'playerBlack': Person
});
```
The only reservation I'd have about implementing this is it may add a fair amount of complexity to the [current](https://github.com/sinclairzx81/typebox/blob/master/src/typebox.ts#L505-L513) schema generation. But id be curious to see an implementation.

Would you be interested in having a go implementing a reference design for this feature? I guess the make or break depends on just how intuitive the schema generation turns out to be for `$ref` types. 

Let me know!

Remirror-zz:
[...]

PR #61, opened 2021-03-24
add $ref support

dwickern:
This PR adds basic $ref support. It works well with ajv and fastify-swagger.

Usage pattern looks something like this:
```ts
// create a schema with unique $id
type Person = Static<typeof Person>
const Person = Type.Object({
    name: Type.String()
}, {
    $id: 'person'
})

fastify.addSchema(Person) // or ajv.addSchema(Person)

fastify.route({
    method: 'GET',
    url: '/',
    schema: {
        response: {
            200: Type.Ref(Person),
        },
    },
    async handler(request, reply): Promise<Person> {
        return { name: 'Derek' }
    },
})
```


src/typebox.ts
```diff
@@ -299,6 +305,14 @@ export class TypeBuilder {
             { ...options, kind: ObjectKind, type: 'object', additionalProperties, properties }
     }
 
+    /** `STANDARD` Creates a `$ref` schema. */
+    public Ref<T extends TObject<any>>(ref: T, options: CustomOptions = {}): TRef<T> {
+        if (!ref.$id) {
+            throw new Error('Referenced schema $id is missing')
```
dwickern, line 311: Ideally `Type.Ref` would require an `$id` defined for its argument.

The problem is that `CustomOptions` (or options in general) are not encoded into the type:
```ts
const Person = Type.Object({
    name: Type.String()
}, {
    $id: 'person'
})

// we should know that Person has an $id defined
const id: string = Person.$id // TS2322: Type 'string | undefined' is not assignable to type 'string'.
```

However that seems like a big change outside the scope of this PR

sinclairzx81:
@dwickern Wow, thanks for the PR! Please allow me some time to review this one. There had actually been some plans for `Type.Box({...})` and `Type.Ref(...)` but had deferred implementation as it required figuring out a larger specification for how collections of schemas should reference and interact as well as downlevel lib space concerns (Your PR is actually the first community mention of how users approach this, and fastify makes for a really good reference usecase, let's discuss!). 

Also, this PR is quite related to https://github.com/sinclairzx81/typebox/issues/34 where I've been pondering things for a while now. I had imagined something along the lines of the following. (which has a few problems with the design but hopefully illustrates the general thinking for `Type.Ref(...)`)

```typescript
const User   = Type.Object({ name: Type.String() })
const Record = Type.Object({ value: Type.String() })
const Box    = Type.Box({ User, Record })

const T = Type.Object({
    user: Type.Ref(Box, 'User'),
    records: Type.Array(
       Type.Ref(Box, 'Record')
    )
})
```
where `T` is
```javascript
const T = {
  definitions: {
    User: {
      $id: '#User',
      type: 'object',
      properties: {
        name: { type: 'string' },
      },
      required: ['name']
    },
    Record: {
      $id: '#Record',
      type: 'object',
      properties': {
        value: { type: 'string' },
      },
      required: ['value']
    }
  },
  type: 'object',
  properties: {
    user: { $ref: '#User' },
    records: { 
         type: 'array',
         items: { $ref: '#Record' }
     }
  }
}
```
The main issue with `$ref` when added in isolation is it ideally needs to be reference into something in such a way as TypeBox can flatten the resulting schema on behalf of the user while still allowing for static type resolution (which could be derived from the keyof in `Type.Box({...})`). Not sure how this design would work in the context of fastify, but would be keen to get your insights. 

@mooyoul Just pinging you on this PR as you had written a `container` type previously and may be able to lend some thoughts.

dwickern:
My intention was to encode the `{ "$ref": "..." }` structure as it appears in JSON Schema. I think it needs to exist in isolation-- consider if the reference is a URI to a different schema file. In that case you might not want to inline the definition. The user needs to make sure the referenced schema exists.

`Box` is interesting idea for a higher-level API that could build on top of this `Ref` primitive.

```ts
const User   = Type.Object({ name: Type.String() })
const Record = Type.Object({ value: Type.String() })
const Box    = Type.Box({ User, Record })

const T = Type.Object({
    user: Box.User, // Ref<User>
    records: Type.Array(
       Box.Record // Ref<Record>
    )
})
```

mooyoul:
@sinclairzx81 Wow. I love the idea of `Box`.  This is way better than my thoughts....
I have a question, How we can optimize outputs if there are a lot of N:1 references?
For example, We have bunch of Schemas that references `User` entity. In this case, how we can remove duplicated `User` definitions in schema?

sinclairzx81:
@dwickern Apoligies for the delay. I've pushed a commit on your branch to enable `$ref` to work across all types, have also removed the the `#` in the `$ref` for `RefKind` (wasn't sure if that was intentional). 

Still mulling over this one. I guess the main bit of uncertainty here is that `Type.Ref()` is actually returning an incomplete schema when used in isolation. For example.

```typescript
const A = Type.Object({ a: Type.String() }, { $id: 'A' })
const B = Type.Object({ a: Type.String() }, { $id: 'B' })
const C = Type.Object({ a: Type.String() }, { $id: 'C' })

const T = Type.Tuple([
   Type.Ref(A),
   Type.Ref(B),
   Type.Ref(C)
])
```
Which results in the following schema. 
```javascript
{
  kind: Symbol(TupleKind),
  type: 'array',
  items: [
    { kind: Symbol(RefKind), '$ref': 'A' }, // Referencing, but where?
    { kind: Symbol(RefKind), '$ref': 'B' },
    { kind: Symbol(RefKind), '$ref': 'C' }
  ],
  additionalItems: false,
  minItems: 3,
  maxItems: 3
}
```
I think this might be a bit unexpected for the user (where the expectation is that all `Type.*` functions return a complete valid schemas). Also it actually assumes a lot of the validation provider too. In the tests provided, you've used `ajv.addSchema()` (which is fine) but not sure if TypeBox should lean on the provider having this capability.

In the past, I'd always imagined `Type.Ref(...)` as being a way to reduce duplication in the output schemas, and primarily for schema publishing scenarios (so not related to AJV at all). With the current `Type.Ref(...)` implementation, because the schema isn't directly consumable, I'm a little apprehensive about introducing it into the library in it's current form. However, one could transform `Type.Ref(...)` into something publishable with the following...

```typescript
const A = Type.Object({ a: Type.String() }, { $id: 'A' })
const B = Type.Object({ a: Type.String() }, { $id: 'B' })
const C = Type.Object({ a: Type.String() }, { $id: 'C' })

const T = Type.Tuple([
   Type.Ref(A),
   Type.Ref(B),
   Type.Ref(C)
])

const U = { definitions: { A, B, C }, ...T }
```
Producing
```javascript
{
  definitions: {
    A: {
      '$id': 'A',
      kind: Symbol(ObjectKind),
      type: 'object',
      additionalProperties: false,
      properties: [Object],
      required: [Array]
    },
    B: {
      '$id': 'B',
      kind: Symbol(ObjectKind),
      type: 'object',
      additionalProperties: false,
      properties: [Object],
      required: [Array]
    },
    C: {
      '$id': 'C',
      kind: Symbol(ObjectKind),
      type: 'object',
      additionalProperties: false,
      properties: [Object],
      required: [Array]
    }
  },
  kind: Symbol(TupleKind),
  type: 'array',
  items: [
    { kind: Symbol(RefKind), '$ref': 'A' },
    { kind: Symbol(RefKind), '$ref': 'B' },
    { kind: Symbol(RefKind), '$ref': 'C' }
  ],
  additionalItems: false,
  minItems: 3,
  maxItems: 3
}
```
This is a little bit more inline with what I'd expect. Although there is a little bit of awkward-ness with respect to the user having to specify the `$id` (which it could be resolve from the `definitions` keys).

I think to proceed, I feel it would be good to iron out the `Type.Box(...)` container type first then release both `Type.Box(...)` and `Type.Ref(...)` as a pair. Just a few quick questions tho,

- What are the benefits of using `ajv.addSchema(...)` when setting up the schematics for a project?
- Wouldn't it just be easier to compile complete (potentially reduced) schemas in AJV? 
- Could this functionality be expressed by just adding `$ref` and `$id` as `CustomOptions`? (so not requiring the new `Type.Ref(...)` type)

Thanks for your patience, look forward to discussing this more.
S

sinclairzx81:
@mooyoul Hi!

> How we can optimize outputs if there are a lot of N:1 references? For example, We have bunch of Schemas that references User entity. In this case, how we can remove duplicated User definitions in schema? 

At this stage, I'm not 100% on the best design here. But I do think something along the lines of what @dwickern is suggesting is probably the way to go, but perhaps limited to referencing into a schema container ONLY. The `Type.Box(...)` type should prevent any duplication.

Perhaps something along the lines of.

```typescript
// domain objects go here.
const Container = Type.Box({
    User: Type.Object({ id: Type.String() })
}, { $id: '/schema/container' })

// dependent schemas go here.
const Foo = Type.Object({
   user: Type.Ref(Container, 'User')
})
const Bar = Type.Object({
    user: Type.Ref(Container, 'User')
})
```
So `Bar` and `Foo` would work as typical TypeBox types. But this would also allow one to publish each schema individually. Something like.. 
```typescript
app.get('/schema/container', (req, res) => res.json(Container))
app.get('/schema/foo', (req, res) => res.json(Foo))
app.get('/schema/bar', (req, res) => res.json(Bar))
```
And where the `Foo` `$ref` is somehow rewritten as the following.
```javascript
{
    type: 'object',
    properties: {
        user: { $ref: '/schema/container#User' }
    }
}
```
Also, to satisfy the case @dwickern is suggesting with this PR, if the following was possible, then I think this would probably be the `Type.Ref(...)` and `Type.Box(...)` spec in a nutshell.

```typescript
// Add all domain schemas in a single call.
ajv.addSchema(Container)
```

@dwickern @mooyoul Open to thoughts around this !

dwickern:
>     Wouldn't it just be easier to compile complete (potentially reduced) schemas in AJV?

For my use-case with fastify, I don't have direct control over the root schema or the top-level `"definitions"`. The framework automatically collects the schema from each route and merges them all into a swagger specification.


>     What are the benefits of using ajv.addSchema(...) when setting up the schematics for a project?

`fastify.addSchema` is the public API for adding a schema with `$id` which you can reference later using `$ref`. Those APIs are described [here](https://www.fastify.io/docs/v2.2.x/Validation-and-Serialization/#adding-a-shared-schema). The argument to `addSchema` must have an `$id`, so you have to add each schema individually.


>     Could this functionality be expressed by just adding $ref and $id as CustomOptions? (so not requiring the new Type.Ref(...) type)

Not exactly. I want the Ref's typescript type (`Static<typeof MyRef>`) to be the same as the type being referenced. I'm using [fastify-typebox](https://github.com/foodsy-app/fastify-typebox) to infer types based on the schema:

```ts
type User = Static<typeof User>
const User = Type.Object({
    name: Type.String()
}, {
    $id: 'user'
})

type NotFoundError = Static<typeof NotFoundError>
const NotFoundError = Type.Object({
    message: Type.String()
})

fastify.addSchema(User)

fastify.route({
    method: 'PUT',
    url: '/:id',
    schema: {
        params: Type.Object({
            id: Type.Integer(),
        }),
        body: Type.Ref(User),
        response: {
            200: Type.Ref(User), // a ref
            404: NotFoundError, // not a ref
        },
    },
    async handler(request, reply) {
        // return type inferred as `Promise<User | NotFoundError>`
        request.params.id // type inferred as `number`
        request.body // type inferred as `User`
    },
})
```

sinclairzx81:
@dwickern Nice addition for the lazy evaluated `Type.Ref(...)`. I've pushed a couple of updates.

- Resolved conflicts from Master
- Update `Type.Ref(...)` to include an `EXPERIMENTAL` comment. 
- Updated the Example script with `Node` example.

I'm actually pretty ok with the API as it stands going out as an `EXPERIMENTAL` feature which could be refined over time. But just need to try and solve the `excessively deep type instantiation` errors. The `Node` example shows where things are popping out at the moment, also the parameter expansions on `Function` and `Constructor` types are popping out and require a `@ts-ignore`. 

The TS recursion limits are a bit of a problem here and I wonder if there is a good solution to this. A little while ago I wrote https://github.com/sinclairzx81/ts-8-bit as an exercise and ran up against recursion limits quite a bit there, but did start to wonder if there is a way to track the recursion depth using a kind of numeric [Iterator](https://github.com/sinclairzx81/ts-8-bit/blob/master/src/ts-8-bit.ts#L55-L67) that would increment on each depth and return `any` if exceeding a maximum depth. Currently TS will just complain and return `any` for the top most object, but it may be possible to do better here (as we know the only way for things to recurse is via `Type.Ref(...)`). Something like the following might be worth exploring.

```typescript
// update to static to include `I` to track depth.
type Static<T extends TSchema, I extends 0 | 1 | 2 | 3> =
   I extends 3 ? any :
   T extends TObject<infer U> ? Static<U, I> : // I
   T extends TRef<infer U> ? Static<U, Next<I>> : // I + 1
   ...
   
type T = Static<typeof T, 0>
```

dwickern:
> But just need to try and solve the excessively deep type instantiation errors.

The function arguments seem to be the problem (more specifically, `...spread`ing them). I implemented a workaround for up to 5 arguments. We could keep it going up to any arity that seems reasonable, or put a catch-all `...any[]` on the last one.

JamieMagee:
If this is working as-is, and you're happy with it going out as an experimental feature, is it possible to merge it and iterate on the design and implementation? I'm very interested in using this to implement the example I gave in #71.

sinclairzx81:
@JamieMagee @dwickern Hey all, yes apologies, still struggling to find time to sit down with this. Immediate thoughts on the work-around; specifically removing variadic tuples to prevent deep instantiation, Is something I really want to take a look at. The current work around is how TypeBox previously handled overloading before variadic tuples were possible in TypeScript. I would like to retain them if possible, mostly because needing to remove them to support `$ref`  strongly hints that TypeBox is pushing at the extents of what's actually possible in TypeScript. Also removing variadics to support `Type.Ref` is a bit of a cost to pay in general, I'd like to find a better solution if one exists.

Will try dig into this over the weekend. Apologies again for the delay.

JamieMagee:
@sinclairzx81 Sorry to ping you on this again, but I was wondering if you've had a chance to think about what an alternative solution might look like?

sinclairzx81:
@dwickern @JamieMagee Hey guys. Apologies for the extended amount of time that has elapsed on this PR. But I finally have an update on this. 

I have just published an experimental version of `$ref` to NPM on `0.17.2` that implements the `Type.Box(...)` and `Type.Ref(...)` approach. This is currently in a provisional state (it needs some other eyes on it other than me), but does satisfy the general use case where I'd want to group related schemas together, and use `$ref` to access them from a set. This is generally the direction I think yields the most natural downstream benefit. 

Documentation for the feature is https://github.com/sinclairzx81/typebox#Reference-Types

The general API is as follows...

```typescript
// Common Types.
const Foo = Type.Object({ a: Type.Number() })
const Bar = Type.String()
const Baz = Type.Array(Type.Number())

// Group these types under a `$id` namespace
const Box = Type.Box("Common", { Foo, Bar, Baz })

// Type.Ref(...) allows for referencing into Box's only.
const T = Type.Object({
   foo: Type.Ref(Box, 'Foo'),
   bar: Type.Ref(Box, 'Bar'),
   baz: Type.Ref(Box, 'Baz'),
})
```
In terms of validation (in the AJV case) you pass the `Box` directly `ajv.addSchema(Box)` which will allow for the dependent type `T` (above) to be validated.
```typescript
const ajv = new Ajv(...)

ajv.addSchema(Box) // add the box to AJV

ajv.validate(T, {
   foo: { a: 1 },
   bar: 'hello',
   baz: []
})
```
Note that this doesn't solve the recursive type problem (that may be addressed at a later time). This update also retains the variadic tuple signatures for `Constructor` and `Function`. 

Happy to leave this PR open for a little while to discuss the implementation. 

Many thanks for all your inputs!
S

sinclairzx81:
@dwickern Hi mate, might look at closing this off this PR in the next day or so. I've held off closing this just in case there's room to offer a `$ref` for non boxed schemas. From your example above, that can be expressed in the following way using the new `Type.Box(...)` and `Type.Ref(...)` functions.

```typescript
const User = Type.Object({ name: Type.String() })
const NotFoundError = Type.Object({ message: Type.String() })
const Model = Type.Box('http://model.app.com', { User, NotFoundError })

fastify.addSchema(Model) // (new) add model schematics to fastify / ajv.

fastify.route({
    method: 'PUT',
    url: '/:id',
    schema: {
        params: Type.Object({
            id: Type.Integer(),
        }),
        body: Type.Ref(Model, 'User'),
        response: {
            200: Type.Ref(Model, 'User'), // a ref
            404: NotFoundError, // not a ref
        },
    },
    async handler(request, reply) {
        // return type inferred as `Promise<User | NotFoundError>`
        request.params.id // type inferred as `number`
        request.body // type inferred as `User`
    },
})
```
For recursive types also mentioned in this thread, I've treated that as a separate method call for `Type.Rec(...)`, that actually allows for a schema to internally reference itself. (it contains a `definitions/self` with a `$ref`). I think doing it that way might for future work should TypeScript provide a solution for recursive type inference in later revisions of the compiler. (hopefully without impacting `$ref` as it's currently implemented).

Note that both implementations are flagged as experimental. There are some considerations there with respect to not expressing the underlying schema as reported by the TS static type (where the referenced types use `$ref` as opposed to the actual properties reported by TypeScript). This also means composition of `Type.Intersect()` needs looking into (as there is no properties to intersect), but will look at this in later revisions.

I feel approaching things this way simplifies the type inference and allows for the variadic tuples to remain for `function` and `constructor` arguments. (I haven't finished with those, still need to figure out where TS is popping out there, it may be a bug with TypeScript as im pretty sure TypeBox isn't pushing recursion limits), But I'm more or less happy with the result.

Thanks for the PR though mate. This ones been a tricky one to reconcile. 
Cheers
S

sinclairzx81:
Closing this PR. Many thanks to everyone for their contributions and insights.

PR #62, opened 2021-03-29
fix clone when options have a null value

dwickern:
If you use `null` in the options, and any method which calls `clone` (Required, Partial, Pick, Omit), you might get an error like this:
```
TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at clone (target/spec/index.js:41:31)
    at /Users/dwickern/code/typebox/target/spec/index.js:42:32
    at Array.reduce (<anonymous>)
    at clone (target/spec/index.js:41:44)
    at /Users/dwickern/code/typebox/target/spec/index.js:42:32
    at Array.reduce (<anonymous>)
    at clone (target/spec/index.js:41:44)
    at /Users/dwickern/code/typebox/target/spec/index.js:42:32
    at Array.reduce (<anonymous>)
    at clone (target/spec/index.js:41:44)
    at TypeBuilder.Omit (target/spec/index.js:190:47)
    at Context.<anonymous> (target/spec/index.js:743:53)
    at processImmediate (internal/timers.js:461:21)
```

This happens because `typeof null === 'object'`.

### DIFF
--- a/spec/schema/omit.ts
+++ b/spec/schema/omit.ts
@@ -11,4 +11,14 @@ describe('Omit', () => {
       const Vector2 = Type.Omit(Vector3, ['z'])
       ok(Vector2, { x: 1, y: 1 })
     })
+  
+    it('User', () => {
+      const User = Type.Object({
+        id: Type.Readonly(Type.Integer()),
+        name: Type.String({ default: null }),
+        email: Type.String({ default: undefined }),
+      });
+      const PartialUser = Type.Omit(User, ['id'])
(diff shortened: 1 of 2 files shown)


sinclairzx81:
Oh hey, thanks for this! :) If only JavaScript provided a built in `Object.clone()` function :P Will merge this one through.

Btw, I'm still mindful of the other `Type.Ref()` PR, but currently struggling trying to find some time to sit down with it (I did want to take a look at how things would line up with a potential `Type.Box()` implementation and provide some feedback). So apologies for the delay. I should be able to take a look over the weekend.  

Thanks again!

PR #63, opened 2021-04-03
feat: emitDecoratorMetadata support

calebboyd:
Typescript emits metadata for "decorated" methods on classes. Specifically, if a class method parameter type is new-able it is added to the `design:paramtypes` metadata created by that flag.

The immediate use case is reflection for automatically scaffolding validators.

eg.
```typescript
class Routes {
	@Route.Post('/')
	route (a: MyTypeBoxType) {
		//here, some router flow has already validated `a` is `MyTypeBoxType`
		//because it was accessible on the `'design:paramtypes'` metadata field
		return doStuff(a)
	}
}
```
---

When using `--emitDecoratorMetadata` with
typescript, the typebox schema will emit as
the JSON value as metadata on the decorated method.
eg.

```typescript
const Qux = Type.Object({
  id: Type.String()
})
type Qux = Static<typeof Qux>
class Foo {
  @foo
  bar(qux: Qux) {
    return qux
  }
}
```

will cause typescript to emit `Qux` instead of `Object`:

```javascript
__metadata('design:paramtypes', [Qux])
```

### DIFF
--- a/example/index.ts
+++ b/example/index.ts
@@ -1,14 +1,34 @@
-import { Type, Static, TObject, TProperties } from '@sinclair/typebox'
+import { Type, Static } from '@sinclair/typebox'
 
-// -----------------------------------------------
-// npm start to run example
-// -----------------------------------------------
+// ------------------------------------------------------
+// decorator
+// ------------------------------------------------------
 
-type Vector = Static<typeof Vector>
-const Vector = Type.Object({
-    x: Type.Number(),
-    y: Type.Number()
-})
+export const method = () => (...args: any) => {}
 
-console.log(Vector)
+// ------------------------------------------------------
+// contract
+// ------------------------------------------------------
 
+type Request = Static<typeof Request>
+const Request = Type.Object({ id: Type.String() })
+
+type Response = Static<typeof Response>
+const Response = Type.Object({ id: Type.String() })
+
+// ------------------------------------------------------
+// service
+// ------------------------------------------------------
+
+export class Service {
+    @method()
+    method(request: Request): Response {
+        return request
+    }
+}
+
+// ...
+
+const service = new Service()
+
+service.method({ id: '1' })
--- a/example/tsconfig.json
+++ b/example/tsconfig.json
@@ -1,7 +1,11 @@
 {
     "compilerOptions": {
         "strict": true,
+        "target": "ESNext",
         "baseUrl": ".",
+        "experimentalDecorators": true,
+        "emitDecoratorMetadata": true,
+        "removeComments": true,
         "paths": {
             "@sinclair/typebox": ["../src/typebox"]
         }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@sinclair/typebox",
-  "version": "0.14.1",
+  "version": "0.16.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@sinclair/typebox",
-      "version": "0.14.1",
+      "version": "0.16.1",
       "license": "MIT",
       "devDependencies": {
         "@types/chai": "^4.1.3",
--- a/src/typebox.ts
+++ b/src/typebox.ts
@@ -101,26 +101,46 @@ export type DictOptions = {
     maxProperties?: number
 } & CustomOptions
 
+/**
+ * This signature causes the type to be emitted by the typescript compiler
+ * when the --emitDecoratorMetadata is specified
+ */
+export type TEmit = {
+    /**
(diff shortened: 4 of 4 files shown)


sinclairzx81:
@calebboyd Hey! Thanks for the PR. This is a pretty neat bit of functionality :D I actually really like it (the `design:*` constructor trick is very cool). Unfortunately, I'm not sure I can accept the PR. The main reason being that it augments all existing `TSchema` types with a non-callable constructor interface. So for example.

```typescript
const Foo = Type.Object({ x: Type.String() })

Foo.apply(1) // TypeScript: Ok!
             // JavaScript: TypeError: Foo.apply is not a function
```
In saying this, I'm actually a huge proponent of stuff like this. So, I've drafted a quick example below of how this could be achieved in a more opt-in way for users without implicating the `TSchema` interface across the board.

**Alternative Approach**

The following code extends the TypeBox `TypeBuilder` class with an additional method `Type.DesignType(...)` that implements the constructor augmentation. The idea here is that down level libraries could extend the `TypeBuilder` with additional methods to help integrate with web service infrastructure (for example, reading type metadata from `__decorator()`). 

```typescript
import { TypeBuilder, TSchema, Static } from '@sinclair/typebox'

// ------------------------------------------------------
// TypeBuilder
// ------------------------------------------------------

export class DesignTypeBuilder extends TypeBuilder {

  public DesignType<T extends TSchema>(schema: T): T & { new (...args: any[]): never } {
    return schema as T & { new (...args: any[]): never }
  }
}

const Type = new DesignTypeBuilder()

// ------------------------------------------------------
// decorator
// ------------------------------------------------------

export const method = () => (...args: any) => {}

// ------------------------------------------------------
// contract
// ------------------------------------------------------

type Request = Static<typeof Request>
const Request = Type.DesignType(Type.Object({ id: Type.String() }))

type Response = Static<typeof Response>
const Response = Type.DesignType(Type.Object({ id: Type.String() }))

// ------------------------------------------------------
// service
// ------------------------------------------------------

export class Service {
    @method()
    method(request: Request): Response {
        return request
    }
}
```
This results in the following generated output
```javascript
const Request = Type.DesignType(Type.Object({ id: Type.String() }));
const Response = Type.DesignType(Type.Object({ id: Type.String() }));

class Service {
  method(request) {
    return request;
  }
}
__decorate([
  exports.method(),
  __metadata("design:type", Function),
  __metadata("design:paramtypes", [Request]),
  __metadata("design:returntype", Response)
], Service.prototype, "method", null);
```

I'm leave this PR open for a few days as I'm happy to discuss this a little more. I'm figuring users of  `NestJS` (and other decorator based web frameworks) would probably get a lot of value from a feature like this. I'm not so sure the feature belongs inside of TypeBox specifically, but at a minimum, perhaps adding some TypeBox documentation around this stuff might be good to start.

Thanks Again!
S

calebboyd:
Thats a much better approach! I had the same concerns with it not actually being a valid constructable...
I take it from your example that `DesignTypeBuilder` is not something you'd want to keep in the project itself? 

If thats the case feel free to close this!

sinclairzx81:
@calebboyd Hey, all good ! Just with respect to the `DesignTypeBuilder` class up there, yeah, that might be something to drop in downstream applications to handle this particular use case, though I'll admit, it's not as nice as just having things work out of the box. I wonder if there is another way to have TS emit on `design` types without the constructor requirement. Hmmm.

That said, TypeBox can be forked and customized downstream (which might be an option for web framework / server developers), but sadly for the main lib, those `constructor` augmentations are a bit too high of a cost to pay :(

Just as a side note though, I do feel this functionality could be a bit of a game changer in terms of how Web Service applications could be written in Node. I actually come from a ASP.NET Web Service / WCF background (a very long time ago), and have been waiting for options to appear to enable a similar code first development flow in Node. So I do think this is worth pursuing in some capacity (would be amazing to see a prototype server framework setup around this stuff). Food for thought.

Thanks for the PR though man. All PR's are super appreciated ! :) 
Cheers
S

calebboyd:
>I do feel this functionality could be a bit of a game changer in terms of how Web Service applications could be written in Node. I actually come from a ASP.NET Web Service / WCF background (a very long time ago), and have been waiting for options to appear to enable a similar code first development flow in Node. So I do think this is worth pursuing in some capacity (would be amazing to see a prototype server framework setup around this stuff). Food for thought.

I feel the same way coming from .net some years ago -- I've been using this strategy for a while with https://github.com/sberan/typed-json-schema/tree/next (may have some inspiration for oneOf and allOf JSON Schema keywords in typebox ;) )
Combined with a [framework](https://github.com/ingress/ingress#getting-started-http) I maintain

PR #64, opened 2021-04-06
Add type keyword for Literal

dwickern:
Set the `"type"` based on the type of the literal being passed.


dwickern:
I thought about using the more specific `integer` type when `Number.isInteger`, but I'm not sure that would really benefit anyone

sinclairzx81:
@dwickern Ah, that's alright, I think `Number.isInteger(1.0)` would probably return `true` in this case anyway. Happy to keep it as `number`. Thank for these PR's :) These are nice refinements!

PR #65, opened 2021-04-06
Add type keyword for KeyOf

dwickern:


sinclairzx81:
Nice, thanks for this :)

PR #66, opened 2021-04-06
Add type keyword for Enum

dwickern:
Derive the `"type"` keyword based on the enum values (either strings, numbers or mixed).

Empty enums are not particularly advisable but it is allowed by the [spec](https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.6.1.2) and included for completeness.


sinclairzx81:
This is an excellent enhancement :) Thanks again!

sinclairzx81:
@dwickern These three PR's have been published on `0.16.2`. Good work !

dwickern:
Thanks for maintaining this project! 🎉

PR #69, opened 2021-04-12
override existing options from a specialized schema

dwickern:
Take the options passed to Omit/Partial/Pick/Required over the options to the original schema.

Fixes this situation:
```ts
const Vector3 = Type.Object({
    x: Type.Number(),
    y: Type.Number(),
    z: Type.Number()
}, {
  title: 'Vector3'
})

const Vector2 = Type.Pick(Vector3, ['x', 'y'], {
  title: 'Vector2'
})

Vector2.title // "Vector3"
```

### DIFF
--- a/spec/schema/omit.ts
+++ b/spec/schema/omit.ts
@@ -1,5 +1,6 @@
 import { Type } from '@sinclair/typebox'
 import { ok, fail } from './validate'
+import { strictEqual } from 'assert'
 
 describe('Omit', () => {
     it('Vector3 to Vector2', () => {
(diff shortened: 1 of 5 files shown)


sinclairzx81:
@dwickern Hey, thanks! :) Happy to merge this one through. 

The original intent here was to try and prevent custom options from overriding known schema properties but the approach taken doesn't work so well for these specialized types. So this is a good catch. I may look into writing some utility function down the line to retain known schema properties, but this will be perfect for now.

Thanks again!

sinclairzx81:
Update published on `0.16.5`

dwickern:
Yeah, a specialized `clone` could match on `kind` to separate known properties from custom options. It would only matter if someone passed a known property as a custom option.

PR #72, opened 2021-04-16
Add Type.Const

wyozi:
Adds `const` from json schema spec. Makes working with discriminated unions nicer


wyozi:
Somehow completely missed Literal, lol

PR #78, opened 2021-07-03
doc: Fix incorrect typescript example for Dict type

khanguslee:
I was looking at the docs on the `README.md` file and found that the typescript type for `Type.Dict` was incorrect. I have amended this.

Also, an alternative way of representing this Dict type would be to do the following:

```ts
type T = Record<string, number>
```

Should we also add this example into the same cell?

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -191,8 +191,8 @@ The following table outlines the TypeBox mappings between TypeScript and JSON sc
 │   	                         │                             │                                │
 ├────────────────────────────────┼─────────────────────────────┼────────────────────────────────┤
 │ const T = Type.Dict(           │ type T = {                  │ const T = {                    │
-│    Type.Number()               │      [key: string]          │    type: 'object'              │
-│ )                              │ } : number                  │    additionalProperties: {     │
+│    Type.Number()               │    [key: string] : number   │    type: 'object'              │
+│ )                              │ }                           │    additionalProperties: {     │
 │   	                         │                             │      type: 'number'            │
 │   	                         │                             │    }                           │
 │   	                         │                             │ }                              │


sinclairzx81:
@khanguslee Hey, thanks :) Just looking at the current `Type.Dict()` documentation, the information there does seem to be correct.
```
├────────────────────────────────┼─────────────────────────────┼────────────────────────────────┤
│ const T = Type.Dict(           │ type T = {                  │ const T = {                    │
│    Type.Number()               │      [key: string]          │    type: 'object'              │
│ )                              │ } : number                  │    additionalProperties: {     │
│   	                         │                             │      type: 'number'            │
│   	                         │                             │    }                           │
│   	                         │                             │ }                              │
│   	                         │                             │                                │
├────────────────────────────────┼─────────────────────────────┼────────────────────────────────┤
```
As for `Record<string, number>` for the TS example, I could be open to that if you feel it helps with readability. In general, `Type.Record()` might actually be a better name than `Type.Dict(...)`. Previously `Type.Dict()` was called `Type.Map()` :D (inline with JavaScripts `Map` object). May consider renaming the method on a subsequent update.

Happy to accept the PR on this. Would it be possible for you to merge in the latest `readme` changes on `sinclairzx81:master` branch first though? (there's been a few edits there recently, and it looks like your version of the readme doesn't have the most recent updates).

Many thanks!

khanguslee:
> @khanguslee Hey, thanks :) Just looking at the current Type.Dict() documentation, the information there does seem to be correct.

I'm confused, are you saying that the docs in `master` is correct still? [Example here](https://www.typescriptlang.org/play?#code/C4TwDgpgBAKgjFAvFA3gKCpqBtA1hEALigGdgAnASwDsBzAXTQF8pjqBXAWwCMJy00oSLABMSVBix4CxMlTr1WUDjz7MBQA).

> As for Record<string, number> for the TS example, I could be open to that if you feel it helps with readability.

Are you using a plugin to generate the table as I am having trouble doing it by hand 😄 .

> May consider renaming the method on a subsequent update.

Would you also be open to having a new section call "Utility Types" in the `README.md` to describe the [utility types](https://www.typescriptlang.org/docs/handbook/utility-types.html) in Typescript? The types added in https://github.com/sinclairzx81/typebox/pull/57 would be suitable candidates to be included in that section.

sinclairzx81:
@khanguslee Hi thanks :)

> I'm confused, are you saying that the docs in master is correct still? Example here.

Lol, whoops yeah, missed that one glancing across it. Good spotting ! :D 

> Are you using a plugin to generate the table as I am having trouble doing it by hand 😄 .

Unfortunately, those tables manually written by hand.  I've considered writing a tool to generate the tables, but it's difficult to get the formatting correct on each cell (and there's minimal space there to actually write in the type). Might make for a nice side project to build out some tool for tabular creation tool for JS data structures, but it would be quite a niche thing. I had wondered if there might be a better way to present these using markdown. Open to suggestions if you have any! :)

> Would you also be open to having a new section call "Utility Types" in the README.md to describe the utility types in Typescript? The types added in #57 would be suitable candidates to be included in that section.

Ah yeah. I had originally grouped `Type.Omit()`, `Type.Pick()`, `Type.Required()` and `Type.Partial()` in a separate group. But figured these types are so often used, they might as well just show in the `Types` list. Happy to leave them there for the time being. 

Thanks again!
S

sinclairzx81:
Ill merge this one in I think. Everything looks good to go. Thanks for the PR! :)

PR #84, opened 2021-07-11
Add "Hidden" modifier

adrian-kriegel:
# TL;DR
I added a Hidden-modifier that removes object properties from the resulting schema but not from the generated types. 
# Usage
```typescript
const T = Type.Object(
{
    a: Type.Hidden(Type.String()),
});
```
Generates: 
```typescript
interface T{
    a?: string;
}
```
```typescript
const T = { 
    type: 'object',
    properties: {},
};
```
# Motivation
The primary motivation for this pr came from REST-APIs. Let's say you have a resource like this: 
```typescript
// could be a tweet or a Facebook post
const post = Type.Object(
{
    content: Type.String(),
    likes: Type.Integer(),
});
```
The resulting schema would contain the property ``likes``. If you used this schema (and nothing else) to validate ``POST`` or ``PUT`` requests, API users would be able to set the number of ``likes`` by updating/creating a post. This can be either solved by manually creating a separate schema or by using the ``Type.Hidden`` modifier on the property ``likes``. 
# Downsides
- hidden types must be object properties
- causes the property to be optional in TypeScript 
- does not use the ``modifier`` property internally

### DIFF
--- a/spec/schema/modifier.ts
+++ b/spec/schema/modifier.ts
@@ -1,5 +1,8 @@
 import { Type, ReadonlyModifier, OptionalModifier } from '@sinclair/typebox'
+
 import * as assert from 'assert'
+
+import { expect } from 'chai';
 
 describe('Modifier', () => {
   it('Omit modifier',  () => {
@@ -7,17 +10,33 @@ describe('Modifier', () => {
     const T = Type.Object({
       a: Type.Readonly(Type.String()),
       b: Type.Optional(Type.String()),
+      c: Type.Hidden(Type.String()),
     })
     
     const S = JSON.stringify(T)
     const P = JSON.parse(S) as any
 
     // check assignment on Type
-    assert.equal(T.properties.a['modifier'], ReadonlyModifier)
-    assert.equal(T.properties.b['modifier'], OptionalModifier)
+    assert.strictEqual(T.properties.a['modifier'], ReadonlyModifier)
+    assert.strictEqual(T.properties.b['modifier'], OptionalModifier)
+
+    expect(T.required).to.be.deep.eq(['a']);
+
+    // Hidden-Modifier
+    expect('c' in T.properties).to.be.false;
     
     // check deserialized
-    assert.equal(P.properties.a['modifier'], undefined)
-    assert.equal(P.properties.b['modifier'], undefined)
-  })
+    assert.strictEqual(P.properties.a['modifier'], undefined)
+    assert.strictEqual(P.properties.b['modifier'], undefined)
+  });
+
+  it('Type.Hidden can only be used on Objects', () => 
+  {
+    const hidden = Type.Hidden(Type.Object({}));
+
+    expect(() => Type.Array(hidden)).to.throw();
+    expect(() => Type.Tuple([Type.String(), hidden])).to.throw();
+    expect(() => Type.Intersect([Type.Object({}), hidden])).to.throw();
+    expect(() => Type.Dict(hidden)).to.throw();
+  });
 })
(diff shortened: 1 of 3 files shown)


sinclairzx81:
@adrian-kriegel Hey thanks for the PR mate. All PR's are super appreciated, but I don't know if i can accept this one.

The main issue with having something like `Type.Hidden()` omit the underlying property, is it has the TypeScript type and schema type diverge. There's been a few cases in the past where I've considered introducing Phantom types that don't match the underlying schema (primarily to assist with inference), but generally steer clear of it as it would cause the static type to report inaccurate information about the underlying schema (which could lead to bugs). So far I've been able to keep these kinds of facades to a minimum, with `Type.Ref()` in particular reporting properties that don't exist (due to the indirection via `$ref`), but structurally and at runtime, the types still line up.

Would any of the following approaches work?

### Optional
Optional might be a good fit here. The property is still present in the schema, but it is flagged as not `required`. TypeScript will flag it as `number | undefined` and JSON schema won't try to validate for it.
```typescript
const Post = Type.Object({
    content: Type.String(),
    likes: Type.Optional(Type.Integer())
})
```
### Omit

Omit will create a new type with the property missing in the schema. I think in this instance you probably want a new type and schema if there is no expectation for the likes to be present across different requests.
```typescript
const Post = Type.Object({
    content: Type.String(),
    likes: Type.Integer()
})
// Omits the likes from the schema, creates a distinct new type (inline with TypeScript)
const PostContent = Type.Omit(Post, ['likes'])
```

### Partial
Partial is pretty common for REST setups, particularly `PATCH`. You can use partial to mark all properties as optional (then selective patch the ones you want in your rest handler)

```typescript
const Post = Type.Object({
    content: Type.String(),
    likes: Type.Integer()
})
// Makes all properties optional
const PartialPost = Type.Partial(Post)
```

### Extend TypeBuilder

If you really need the property to vanish, it may be possible to extend TypeBox' `TypeBuilder`.  There is a great project making use of this functionality [here](https://github.com/useflyyer/flyyer-variables/blob/master/src/index.ts) which takes things quite far.
```typescript
import { TypeBuilder, TSchema } from '@sinclair/typebox'

export class TypeBuilderExtended extends TypeBuilder {
    public Hidden<T extends TSchema>(t: T): T {
         // Implement facade here
         
         // @ts-ignore
         return undefined
    }
}

export const Type = new TypeBuilderExtended()
```
 
Thanks mate, will leave the PR open for a bit, but have a quick review of the above and see if they could fit your usecase.

Cheers
S

adrian-kriegel:
Hi, first of all thank you for maintaining this project so actively. I absolutely get your point deciding not to include it. Just in case anyone else needs this kind of thing and stumbles upon this thread, I would like to discuss your proposed solutions. 
### Partial / Optional
I think using ``Partial`` and ``Optional`` more or less boils down to the same thing. You would still be able to write to the protected field though the API. 
### Omit
Using ``Omit`` is probably the best solution, although you absolutely need to make sure you name all the protected properties.
### Extend TypeBuilder
Extending the TypeBuilder is what I did in the PR but the way you implemented it will not work with your ``Type.Object`` as the hidden property would be undefined but still be listed in ``Object.keys(...)`` causing a ``TypeError``. Instead, ``Type.Object`` would need to filter out all ``undefined`` properties. I initially implemented it using the ``modifier`` property, which is why the approach in the PR slightly differs from this approach.

sinclairzx81:
@adrian-kriegel Thanks mate :) May look into adding linter at a later time. For now I'm fairly happy manually lining up and structuring source code for readability over convention. Those conditional and mapped types are somewhat challenging to read as is :D 

Thanks again mate, will close off this one.
S

PR #109, opened 2021-10-06
Add Date type

toby-wollit:


toby-wollit:
Hi! We use binary serialisation (notepack) for communication with our api which supports Date objects. I agree that it doesn't make sense for JSON serialisation.

Perhaps another option would be allowing users to somehow extend TSchema with custom types which would allow us to add Date internally.

sinclairzx81:
Yup, that should work. `Type.Facade(...)` accepts `CustomOptions` like the other types, so you should be able to pass anything in there. So for example.

```typescript
const T = Type.Facade<Date>({ type: 'date' }) // const T = { type: 'date' }
```
And I guess in theory, it should also be possible to construct a schema manually (I guess in instances where users want fine level control over the output schema)

```typescript
export interface Foo {
    value: string
}

const T = Type.Facade<Foo>({
   type: 'object',
   properties: {
       value: { type: 'string' }
   }
})

type T = Static<typeof T> // type T = Foo
```
It's probably this latter example that yields the most benefit to TypeBox, it may also help Open API users create specific schemas. I'll give this some thought over the next few days.

Cheers

toby-wollit:
Awesome! that would be excellent, thank you :)

panzelva:
@sinclairzx81 
Hello - first of all, thanks for this great library ;) We are using it for multiple project and its mostly a blast.

However there is one thing missing for us and it would actually be solved by the `Facade` example that you posted here.

Are there any plans to bring this functionality to the library?

sinclairzx81:
@panzelva Hi. There has been a little bit of work done on this recently. It is now possible to create your own arbitrary custom types so long as they extend `TSchema` and define a typed phantom `$static` property which TypeBox uses to inference the type. With this you can create your own `Facade<T>` type, consider the following.

TypeScript Playground Link [Here](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgFQJ5gKYBo4GUYCGMwAxjsriQBYYgFwC+cAZlBCHAOQACAzsADsSAGwLAoAehjoMAIwgAPTgFgAUGsEwMUZgRIYUAMT0EAJhgA8yAHxwMCrQNO8UlGnURq4cACS9CxCQAXChqDGpqzACuQsQQAnDGJGaWNgAUAG4EwlEYITEA1gIQAO4CAJSIcFAYMFFQCVk5BgQuyEkpVrbh6qok8f4ocAC8KDIAdADysgBWGCQwaQhecAQhHeYW-lCCAObWS3DSmCGczBAQnIzlWCuy6yabAlEgstoHSMd5XLIEUFcMG4rYKJR6WeQQYQYAgCD5HGSnX4ALwB5TCaN6XyGo3wRFIFi+EGYKGsagkEnhmGxnlU5O8qxC2z2ZIp3nucGer20LPpIIhUJhPJ6aiAA)

```typescript
import { Type, Static, TSchema } from '@sinclair/typebox'

interface TFacade<T> extends TSchema {
  $static: T
}

function Facade<T>(value: unknown) { return value as TFacade<T> }

const T = Type.Object({
  a: Facade<string>({ type: 'foo' }),
  b: Facade<number>({ type: 'bar' }),
  c: Facade<boolean>({ type: 'baz' })
})

type T = Static<typeof T>
// type T = {
//   a: string
//   b: number
//   c: boolean
// }
```
This functionality was added recently to try provide more options for users who need custom schemas above and beyond the standard TypeBox schemas. It was also part of some significant optimization work for TypeScript 4.5. 

Hope this helps
S

PR #118, opened 2021-11-12
fix: Type.Optional() set `nullable: true`

Divlo:
Hello! :wave: @sinclairzx81 

## Current Behavior

Currently, a `null` value of `type: "string"` in the `ajv` schema is serialized as a double-quoted empty string (`""`)

## Expected Behavior

A  `null` string should be serialized as `null`, and not `""`.

## Solution

This PR adds `nullable: true` to the schema object that returns `Type.Optional`.
Might be a **BREAKING CHANGE**, I don't really know to be honest but, maybe we could add an option to `Type.Optional` to not add this behavior by default.

## References

- https://ajv.js.org/json-schema.html#nullable
- https://github.com/fastify/fast-json-stringify/issues/152

### DIFF
--- a/src/typebox.ts
+++ b/src/typebox.ts
@@ -36,7 +36,7 @@ export const ReadonlyModifier                    = Symbol('ReadonlyModifier')
 
 export type TModifier                            = TReadonlyOptional<TSchema> | TOptional<TSchema> | TReadonly<TSchema>
 export type TReadonlyOptional<T extends TSchema> = T & { modifier: typeof ReadonlyOptionalModifier }
-export type TOptional<T extends TSchema>         = T & { modifier: typeof OptionalModifier }
+export type TOptional<T extends TSchema>         = T & { modifier: typeof OptionalModifier, nullable?: boolean }
 export type TReadonly<T extends TSchema>         = T & { modifier: typeof ReadonlyModifier }
 
 // ------------------------------------------------------------------------
@@ -285,7 +285,10 @@ export class TypeBuilder {
 
     /** `STANDARD` Modifies a schema object property to be `optional`. */
     public Optional<T extends TSchema>(item: T): TOptional<T> {
-        return { ...item, nullable: true, modifier: OptionalModifier }
+        if (item.type === 'string') {
+            return { ...item, nullable: true, modifier: OptionalModifier }
+        }
+        return { ...item, modifier: OptionalModifier }
     }
 
     /** `STANDARD` Creates a Tuple schema. */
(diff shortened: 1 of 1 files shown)


sinclairzx81:
@Divlo Hi, thanks for the PR. Unfortunately I can't merge this one as `nullable` isn't a valid JSON schema keyword. The `nullable` keyword is only present in the Open API / Swagger specifications, and I believe they may be dropping support for it and moving to draft `2020-12` onwards. Refer to relevant issues below.

https://github.com/json-schema-org/json-schema-spec/issues/1067
https://github.com/OAI/OpenAPI-Specification/pull/2246

## Optional Properties

The `Type.Optional(T)` modifier is used to make a object property `T | undefined`. It mirrors the TypeScript `prop?: T` syntax for specifying the potential absence of value. 

Consider the following.

```typescript
// TypeScript
interface Foo {
   name?: string // string | undefined
}
// TypeBox
const Foo = Type.Object({
   name: Type.Optional(Type.String()) // string | undefined
})
// JSON Schema
const Foo = { 
  type: 'object', 
  properties: { 
     name: { type: 'string' } 
  },
  required: []  // not required
}
```
## Handling Null

TypeBox handles null in the same way as TypeScript, where null is treated as a value type similar to number or string. TypeBox uses union types to describe values that can be either null or some other type.

Consider the following

```typescript
// TypeScript
interface Foo {
    name: string | null
}
// TypeBox
const Foo = Type.Object({
   name: Type.Union([Type.String(), Type.Null()]) // string | null
})
// JSON Schema
const Foo = { 
  type: 'object', 
  properties: { 
    name: { 
      anyOf: [
        { type: 'string' },
        { type: 'null' }
      ]
    } 
  },
  required: ['name']  // required
}
```

##  Implementing Nullable

TypeBox doesn't provide out of the box support for the `nullable` keyword, but does allow users to construct custom types that map back to TypeBox's Type System. The following implements `Type.Nullable(T)` using generics.

```typescript
function Nullable<T extends TSchema>(schema: T): TUnion<[T, TNull]> {
    return { ...schema, nullable: true } as any
}

const T = Nullable(Type.String())        // const T = {
                                         //   type: 'string',
                                         //   nullable: true
                                         // }

type T = Static<typeof T>                // type T = string | null

```
For additional information, refer to the (OpenAPI)[https://github.com/sinclairzx81/typebox#openapi] section of the TypeBox readme.

Divlo:
Thanks for your detailed explanation! @sinclairzx81 

>I believe they may be dropping support for it and moving to draft 2020-12 onwards. Refer to relevant issues below.

Indeed, we should not implement things that are not supported anymore, I didn't know that `nullable` was only present in the Open API / Swagger specifications. :+1:

This solved the problem:
```ts
const Foo = Type.Object({
   name: Type.Union([Type.String({ minLength: 1 }), Type.Null()]) // string | null
})
```

Set `minLength: 1`, allows to convert `""` to `null` and thanks to `Type.Union`, `null` value stays `null`. :+1: 

Closing, as the issue is solved (I should have opened an issue before opening a PR my bad).

Issue #129, opened 2021-11-29
help: Changes since 0.14 to 0.22 regarding union, null, and not-required properties


thatsnotright:
We are upgrading, and have this helper defined:

```
export function NotRequired<T extends TSchema>(otherType: T): TOptional<TUnion<[TNull, T]>> {
  return Type.Optional(Type.Union([Type.Null(), otherType]))
}
```

which allowed the property to be missing, null, or the provided type.

Now I cannot get a schema generated that does not have the field as required, whereas before it used to be omitted from the 'required' array.

This
```
export const CONTAINER_TYPE = Type.Intersect([
  BASE_ENTITY,
  Type.Object({
    name: Type.String(),
    maxChildren: NotRequired(Type.Number()),
    isDefaultContainer: Type.Optional(Type.Boolean()),
    width: NotRequired(Type.Number()),
    layout: NotRequired(Type.String()),
    maxVolume: NotRequired(
      Type.Object({
        unit: EnumType(Volume),
        value: Type.Number(),
      })
    ),
    maxMass: NotRequired(
      Type.Object({
        unit: EnumType(Mass),
        value: Type.Number(),
      })
    ),
  }),
])
```

only works if I add an additional `Type.Optional` around the call to `NotRequired` even though it also has also a `Type.Optional` and inspecting the result includes the `Symbol('OptionalModifier')` as the TSchema `modifier` field.

I added some debug code to the library, and basically it comes down to Symbol('OptionalModifier') being different between the comparison in `Object()` for `exports.OptionalModifier` and the candidate's modifier.  Even though they're both visually identical.

The helper function we have above is declared in a common library package (we have a monorepo with npm workspaces), whereas adding the Type.Optional in the project directly works.  I assume this is some kind of issue with using Symbol between the node_module in the common package file, and the node_module in the server project?  I'm at a loss how to fix this, curious if you have any ideas what we may be doing that's now triggering this behavior?  Thank you!

thatsnotright:
I assume this is the reason https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#shared_symbols_in_the_global_symbol_registry

(though we are using typebox in Node)

sinclairzx81:
@thatsnotright Hi, the `Type.Optional(...)` modifier is actually the correct way to omit a property from the schemas `required` array. So for example, consider the `width` property defined on the `CONTAINER_TYPE` with and without the `Type.Optional(...)` modifier.

[...]

PR #130, opened 2021-11-29
fix: Utilize the same symbol across realms

thatsnotright:
This will convert the Symbol's to look up an existing global symbol, if it's available.

I verified locally that using Symbol.for resolves my issue #129 but there may be side-effects that I do not intend for other projects which use multiple versions of typebox.


sinclairzx81:
Closing. See https://github.com/sinclairzx81/typebox/issues/129#issuecomment-991519021

PR #145, opened 2022-01-22
[WIP, but already worth reviewing] Implement brand types

castarco:
This PR introduces brand types for string literals. It could have been a bit bigger and do the same for numbers, but I preferred to start with strings as it is usually this type the one deserving more attention.

Another goal of this PR is to keep compatibility with the package `@coderspirit/nominal` ( https://github.com/Coder-Spirit/nominal ). It doesn't add it as a dependency, but in exchange it depends on a simpler shared dependency (`@coderspirit/nominal-symbols`).

From what I've seen in TypeBox's code, it seems like it was developed trying to minimise the amount of dependencies, so I feel obliged to explain my decision:
  - `@coderspirit/nominal` relies on "unique symbols" to implement the typical branding stuff, in part to make it more difficult to forge types, but more importantly, in order to avoid leaking stuff by accident through external APIs and minimising potential side effects at runtime.
  - The problem with "unique symbols" is that they are, indeed, unique. This means that if different packages choose different branding strategies, it gets more difficult to assert that two types are the "same".
  - Regarding why `@coderspirit/nominal` needs "special treatment" to keep compatibility, that's because it goes a bit further than just providing nominal typing (It allows to use some weak forms of dependent typing that, for example, can help to perform taint analysis via the type system).
  
## Summarised goals:
  - implement type branding for string values
  - ensure compatibility with `@coderspirit/nominal`
  - related to the previous goal: I intend to propose the same kind of change to other libraries such as MikroORM, to enhance type safety across the whole backend stack, from external APIs towards the persistence layer.

## Regarding Deno support
- For now I released a Deno-compatible version of `@coderspirit/nominal`, ~but not yet a Deno-compatible version of `@coderspirit/nominal-symbols` (which I'll do, I guess that this PR will have to wait until that to be mergeable)~ (this point is already fixed).
- What I'm not sure about is, how to import the symbols in a way that the code is compatible with Node and Deno at the same time... 🤔 (I'm applying some dirty tricks for that in `@coderspirit/nominal`, but I don't want to impose that toll on packages I don't maintain myself).

I'll add more specific comments next to the modified code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,9 @@
     "spec:schemas": "hammer task spec_schemas",
     "test": "npm run spec"
   },
+  "dependencies": {
+    "@coderspirit/nominal-symbols": "^1.0.1"
+  },
   "devDependencies": {
     "@sinclair/hammer": "^0.15.8",
     "@types/chai": "^4.3.0",
--- a/package.json
+++ b/package.json
@@ -27,14 +27,14 @@
   },
   "devDependencies": {
     "@sinclair/hammer": "^0.15.8",
-    "@types/chai": "^4.2.22",
-    "@types/mocha": "^9.0.0",
+    "@types/chai": "^4.3.0",
+    "@types/mocha": "^9.1.0",
     "@types/node": "^16.11.9",
-    "ajv": "^8.8.2",
+    "ajv": "^8.9.0",
     "ajv-formats": "^2.1.1",
     "chai": "^4.3.4",
-    "mocha": "^9.1.3",
-    "tsd": "^0.19.0",
-    "typescript": "^4.5.2"
+    "mocha": "^9.1.4",
+    "tsd": "^0.19.1",
+    "typescript": "^4.5.5"
   }
 }
--- a/spec/types/branded-string.ts
+++ b/spec/types/branded-string.ts
@@ -1,0 +1,10 @@
+import * as Spec from './spec'
+import { Brand, Type } from './typebox'
+
+type Email = Brand<string, 'Email'>
+type WrongEmail = Brand<string, 'WrongEmail'>
+
+// For branded types we require to pass extra constraints, otherwise it would
+// make no sense to use them at all.
+Spec.expectType<Email>(Spec.infer(Type.BrandedString<Email>({ format: 'email' })))
+Spec.expectNotType<Email>(Spec.infer(Type.BrandedString<WrongEmail>({ format: 'email' })))
--- a/spec/types/branded-string.ts
+++ b/spec/types/branded-string.ts
@@ -7,4 +7,7 @@ type WrongEmail = Brand<string, 'WrongEmail'>
 // For branded types we require to pass extra constraints, otherwise it would
 // make no sense to use them at all.
 Spec.expectType<Email>(Spec.infer(Type.BrandedString<Email>({ format: 'email' })))
+
+// We validate that brands clearly distinguish between different types.
 Spec.expectNotType<Email>(Spec.infer(Type.BrandedString<WrongEmail>({ format: 'email' })))
+Spec.expectNotType<Email>(Spec.infer(Type.String()))
--- a/src/typebox.ts
+++ b/src/typebox.ts
@@ -26,6 +26,8 @@ THE SOFTWARE.
 
 ---------------------------------------------------------------------------*/
 
+import { __BaseType, __Brand } from '@coderspirit/nominal-symbols'
+
 // --------------------------------------------------------------------------
 // Modifiers
 // --------------------------------------------------------------------------
@@ -38,6 +40,20 @@ export type TModifier                            = TReadonlyOptional<TSchema> | 
 export type TReadonlyOptional<T extends TSchema> = T & { modifier: typeof ReadonlyOptionalModifier }
 export type TOptional<T extends TSchema>         = T & { modifier: typeof OptionalModifier }
 export type TReadonly<T extends TSchema>         = T & { modifier: typeof ReadonlyModifier }
+
+// --------------------------------------------------------------------------
+// Nominal Types
+// --------------------------------------------------------------------------
+export type BrandType                                 = string | symbol
+export type Brand<BaseType, Brand extends BrandType>  = BaseType & {
+  readonly [__BaseType]: BaseType
+  readonly [__Brand]: Brand
+}
+type Flavor<BaseType, Brand extends BrandType>        = BaseType & {
+  readonly [__BaseType]?: BaseType | undefined
+  readonly [__Brand]?: Brand | undefined
+}
+type StringFlavor                                     = Flavor<string, BrandType>
 
 // --------------------------------------------------------------------------
 // Schema Standard
@@ -81,7 +97,7 @@ export type StringFormatOption =
     | 'uuid'         | 'iri-reference' | 'uri-template' | 'json-pointer' | 'relative-json-pointer'
     | 'regex'
 
-export declare type StringOptions<TFormat extends string> = {
+export declare type StringOptions<TFormat extends string | undefined = undefined> = {
     minLength?: number
     maxLength?: number
     pattern?: string
@@ -146,7 +162,7 @@ export interface TArray     <T extends TSchema>                       extends TS
 export interface TLiteral   <T extends TValue>                        extends TSchema, CustomOptions         { $static: StaticLiteral<T>,   kind: typeof LiteralKind,   const: T }
 export interface TEnum      <T extends TEnumKey[]>                    extends TSchema, CustomOptions         { $static: StaticEnum<T>,      kind: typeof EnumKind,      anyOf: T }
 export interface TRef       <T extends TSchema>                       extends TSchema, CustomOptions         { $static: Static<T>,          kind: typeof RefKind,       $ref: string  }
-export interface TString                                              extends TSchema, StringOptions<string> { $static: string,             kind: typeof StringKind,    type: 'string' }
+export interface TString    <TBrand extends StringFlavor = string>    extends TSchema, StringOptions<string> { $static: TBrand,             kind: typeof StringKind,    type: 'string' }
 export interface TNumber                                              extends TSchema, NumberOptions         { $static: number,             kind: typeof NumberKind,    type: 'number' }
 export interface TInteger                                             extends TSchema, NumberOptions         { $static: number,             kind: typeof IntegerKind,   type: 'integer' } 
 export interface TBoolean                                             extends TSchema, CustomOptions         { $static: boolean,            kind: typeof BooleanKind,   type: 'boolean' } 
@@ -325,6 +341,11 @@ export class TypeBuilder {
 
     /** `Standard` Creates a string type */
     public String<TCustomFormatOption extends string>(options: StringOptions<StringFormatOption | TCustomFormatOption> = {}): TString {
+        return this.Store({ ...options, kind: StringKind, type: 'string' })
+    }
+
+    /** Like string, but "branded". It enforces passing options */
+    public BrandedString<TBrand extends StringFlavor, TCustomFormatOption extends string | undefined = undefined>(options: StringOptions<StringFormatOption | TCustomFormatOption>): TString<TBrand> {
         return this.Store({ ...options, kind: StringKind, type: 'string' })
     }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,6 +8,9 @@
       "name": "@sinclair/typebox",
       "version": "0.23.2",
       "license": "MIT",
+      "dependencies": {
+        "@coderspirit/nominal-symbols": "^1.0.1"
+      },
       "devDependencies": {
         "@sinclair/hammer": "^0.15.8",
         "@types/chai": "^4.3.0",
@@ -125,6 +128,17 @@
       },
       "engines": {
         "node": ">=4"
+      }
+    },
+    "node_modules/@coderspirit/nominal-symbols": {
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/@coderspirit/nominal-symbols/-/nominal-symbols-1.0.1.tgz",
+      "integrity": "sha512-Vlb0AgN7aciQ1W+gWzejr+cewFt1xSmDmBm7lO5uuNQ3AAzRuQoOPGrROToTsoVMbq5W+D7cj8O9XqUpOIP18g==",
+      "engines": {
+        "node": ">=12.0.0"
+      },
+      "peerDependencies": {
+        "typescript": ">=4.0"
       }
     },
     "node_modules/@nodelib/fs.scandir": {
@@ -2009,7 +2023,6 @@
       "version": "4.5.5",
       "resolved": "https://registry.npmjs.org/typescript/-/typescript-4.5.5.tgz",
       "integrity": "sha512-TCTIul70LyWe6IJWT8QSYeA54WQe8EjQFU4wY52Fasj5UKx88LNYKCgBEHcOMOrFF1rKGbD8v/xcNWVUq9SymA==",
-      "dev": true,
       "bin": {
         "tsc": "bin/tsc",
         "tsserver": "bin/tsserver"
@@ -2268,6 +2281,12 @@
           }
         }
       }
+    },
+    "@coderspirit/nominal-symbols": {
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/@coderspirit/nominal-symbols/-/nominal-symbols-1.0.1.tgz",
+      "integrity": "sha512-Vlb0AgN7aciQ1W+gWzejr+cewFt1xSmDmBm7lO5uuNQ3AAzRuQoOPGrROToTsoVMbq5W+D7cj8O9XqUpOIP18g==",
+      "requires": {}
     },
     "@nodelib/fs.scandir": {
       "version": "2.1.5",
@@ -3638,8 +3657,7 @@
     "typescript": {
       "version": "4.5.5",
       "resolved": "https://registry.npmjs.org/typescript/-/typescript-4.5.5.tgz",
-      "integrity": "sha512-TCTIul70LyWe6IJWT8QSYeA54WQe8EjQFU4wY52Fasj5UKx88LNYKCgBEHcOMOrFF1rKGbD8v/xcNWVUq9SymA==",
-      "dev": true
+      "integrity": "sha512-TCTIul70LyWe6IJWT8QSYeA54WQe8EjQFU4wY52Fasj5UKx88LNYKCgBEHcOMOrFF1rKGbD8v/xcNWVUq9SymA=="
     },
     "uri-js": {
       "version": "4.4.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,54 +1,54 @@
 {
   "name": "@sinclair/typebox",
-  "version": "0.22.0",
+  "version": "0.23.2",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@sinclair/typebox",
-      "version": "0.22.0",
+      "version": "0.23.2",
       "license": "MIT",
       "devDependencies": {
         "@sinclair/hammer": "^0.15.8",
-        "@types/chai": "^4.2.22",
-        "@types/mocha": "^9.0.0",
+        "@types/chai": "^4.3.0",
+        "@types/mocha": "^9.1.0",
         "@types/node": "^16.11.9",
-        "ajv": "^8.8.2",
+        "ajv": "^8.9.0",
         "ajv-formats": "^2.1.1",
         "chai": "^4.3.4",
-        "mocha": "^9.1.3",
-        "tsd": "^0.19.0",
-        "typescript": "^4.5.2"
+        "mocha": "^9.1.4",
+        "tsd": "^0.19.1",
+        "typescript": "^4.5.5"
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.16.0",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.16.0.tgz",
-      "integrity": "sha512-IF4EOMEV+bfYwOmNxGzSnjR2EmQod7f1UXOpZM3l4i4o4QNwzjtJAu/HxdjHq0aYBvdqMuQEY1eg0nqW9ZPORA==",
+      "version": "7.16.7",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.16.7.tgz",
+      "integrity": "sha512-iAXqUn8IIeBTNd72xsFlgaXHkMBMt6y4HJp1tIaK465CWLT/fG1aqB7ykr95gHHmlBdGbFeWWfyB4NJJ0nmeIg==",
       "dev": true,
       "dependencies": {
-        "@babel/highlight": "^7.16.0"
+        "@babel/highlight": "^7.16.7"
       },
       "engines": {
         "node": ">=6.9.0"
       }
     },
     "node_modules/@babel/helper-validator-identifier": {
-      "version": "7.15.7",
-      "resolved": "https://registry.npmjs.org/@babel/helper-validator-identifier/-/helper-validator-identifier-7.15.7.tgz",
-      "integrity": "sha512-K4JvCtQqad9OY2+yTU8w+E82ywk/fe+ELNlt1G8z3bVGlZfn/hOcQQsUhGhW/N+tb3fxK800wLtKOE/aM0m72w==",
+      "version": "7.16.7",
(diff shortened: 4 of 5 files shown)


package.json
```diff
@@ -25,16 +25,19 @@
     "spec:schemas": "hammer task spec_schemas",
     "test": "npm run spec"
   },
+  "dependencies": {
+    "@coderspirit/nominal-symbols": "^1.0.1"
```
castarco, line 29: I could not avoid adding this dependency given the goals of this PR.

spec/types/branded-string.ts
```diff
@@ -0,0 +1,13 @@
...
+type Email = Brand<string, 'Email'>
+type WrongEmail = Brand<string, 'WrongEmail'>
+
+// For branded types we require to pass extra constraints, otherwise it would
+// make no sense to use them at all.
+Spec.expectType<Email>(Spec.infer(Type.BrandedString<Email>({ format: 'email' })))
+
+// We validate that brands clearly distinguish between different types.
+Spec.expectNotType<Email>(Spec.infer(Type.BrandedString<WrongEmail>({ format: 'email' })))
+Spec.expectNotType<Email>(Spec.infer(Type.String()))
```
castarco, line 13: This showcases the introduced feature:
  - brand strings only work when we pass "configuration" to `BrandedString` (otherwise it would make no sense to attach brands to those values)
  - I had to use a different function than `String` to avoid breaking backwards compatibility. The name is completely debatable, it doesn't have to be `BrandedString`, I'm open to suggestions.

src/typebox.ts
```diff
@@ -39,6 +41,20 @@ export type TReadonlyOptional<T extends TSchema> = T & { modifier: typeof Readon
...
+export type BrandType                                 = string | symbol
+export type Brand<BaseType, Brand extends BrandType>  = BaseType & {
+  readonly [__BaseType]: BaseType
+  readonly [__Brand]: Brand
+}
+type Flavor<BaseType, Brand extends BrandType>        = BaseType & {
+  readonly [__BaseType]?: BaseType | undefined
+  readonly [__Brand]?: Brand | undefined
+}
+type StringFlavor                                     = Flavor<string, BrandType>
```
castarco, line 56: We don't export flavors because, in the context of TypeBox, there's no use on making them accessible. We need them, though, to ease the implementation of internal details.

As before, I'm completely open to discuss the names of these types.

src/typebox.ts
```diff
@@ -81,7 +97,7 @@ export type StringFormatOption =
     | 'uuid'         | 'iri-reference' | 'uri-template' | 'json-pointer' | 'relative-json-pointer'
     | 'regex'
 
-export declare type StringOptions<TFormat extends string> = {
+export declare type StringOptions<TFormat extends string | undefined = undefined> = {
```
castarco, line 100: As the `format` property was optional, there's no harm on extending the `TFormat` type constrain, and to give it a default value. This is actually useful when defining the `BrandedString` function.

src/typebox.ts
```diff
@@ -26,6 +26,8 @@ THE SOFTWARE.
 
 ---------------------------------------------------------------------------*/
 
+import { __BaseType, __Brand } from '@coderspirit/nominal-symbols'
```
castarco, line 29: For compatibility with Deno, this line should be replaced by, this requires post-processing 😢 .
```typescript
import { __BaseType, __Brand } from 'https://deno.land/x/nominal_symbols@1.0.2/nominal-symbols/deno/index.ts'
```

sinclairzx81:
Hi. Thanks for this PR. Unfortunately I'm not sure I can accept these updates as they run a little counter to some of the design goals of the project. These goals are to align specifically to the TS structural type system (and how it reasons about types), contrasted with JSON schema representations of those types.

In TypeScript, nominal typing is typically represented today as a composite type, usually expressed as `T & { __brand: '...' }` which allows the structural type system to discern one type over another. However current approaches are typically limited to simple primitives (strings and numbers). I think the expectation would be for the TS language to be able to reason naturally about classes and objects also, as well as understand sub classed types and their relationships. It's because TypeScript is still very much focused on structural typing (with patterns for nominal typing essentially a workaround), it's this that makes me a bit apprehensive on introducing facilities for nominal types in TypeBox.

There is a long standing issue dating back to 2014 [here](https://github.com/Microsoft/TypeScript/issues/202) which requests support for nominal typing (essentially hinting to TypeScript that it should treat a type as nominal). I think if such a feature lands in TypeScript I would almost certainly look at implementing this as `Type.Nominal(...)` (to analog a syntax modifier hint), but the requirement here would be that TS could naturally reason about the type (and it's relationships), and there were ways to express such a type in JSON schema. (sub classing would be interesting in this regard)

With this said, because current patterns for nominal types are currently expressed as  `T & { __brand: '...' }` composites, it is actually possible to express these types with TypeBox. There are a couple of ways to achieve this, as below.

### Custom Type

The following creates a `TNominal` type which augments the TS representation, but omits the `__brand` from the schema representation. This would be the recommended way to express a custom type of this nature where the `__brand` would only be meaningful to TS, not to JSON schema.

Playground Link [Here](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgFQJ5gKYBo4GUYCGMwAxjsriQBYYgFwC+cAZlBCHAOQACAzsADsSAGwLAoAehjoMAIwgAPTgFgAUGokS4AWl179Bw0eMnTRjVuQA5doILCAPMhwAhKAQEATAHwWdZgMCgszU1DAVIWDhBGAwoZgISDBQbEDtHZDhw2K9eFEoaOld3LyyFHM883hgoQQBzbzKKvIpqWnoENTg4ABJqolIALhQAbU4+wmISTgBdOAAyRDgAfWXZEs9htw9PRjUGUNVmAFchYggBOFT0pyaMXPy2orht0uz7yrhq2oEGgApeE8CMNnCs1hsthsAJQg64CexOYo7RqdVTdKAYGDHKCXJAAOgJgMK9CYBDyHlQcE0cFsMGGq3WOxYbA4RPa+0O1OC3J5QT8AFEFARwMIMH5eRLJQZDiQLtU4AA1ezATxEDCefA-OpwAC8V1s8OEfzQmDxmvqfyhOE4SuEKrVGpq9U4UMO0kwiuVqtijq1urwk1IDndGAgzE9du96vNv186iOpxI50uADcvWq-mnhMcMMNvvUYRH7T6Y9rUejMdjU-Yc3AyUWo776hz4yczsALnAwD8YJma7mGw7S1Clgd492Yn3IxnODRhMIIC6R9SIABrQ4TgS92cYeeLkfdQ9HqlaBLAYRAA)

```typescript
import { Type, Static, TSchema } from '@sinclair/typebox'

// --------------------------------------------------------------
// TNominal<T, Brand>
// --------------------------------------------------------------

export interface TNominal<T extends TSchema, Brand extends string> extends TSchema {
  $static: T['$static'] & { __brand: Brand }
}

function Nominal<T extends TSchema, Brand extends string>(schema: T, __brand: Brand): TNominal<T, Brand> {
  return { ...schema } as any // omit: __brand from schema
}

// --------------------------------------------------------------
// Example
// --------------------------------------------------------------

const ValidatedString = Nominal(Type.String(), 'ValidatedString')

type ValidatedString = Static<typeof ValidatedString>

function validate(value: string): ValidatedString {
  return value as ValidatedString
}

function print(value: ValidatedString) { }

print(validate('hello')) // ok

print('hello')           // fail

```
### Schema Type

The following creates the same Nominal type, however the `__brand` is included in the output schema. This unfortunately results in the output schema being non-valid as JSON schema can't intersect both `string` and `object` via `allOf`.

Playground Link [Here](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgFQJ5gKYBo4GUYCGMwAxjsriQBYYgFwC+cAZlBCHAOQACAzsADsSAGwLAoAehjoMAIwgAPTgFgAUGokS4AWl179Bw0eMnTujVoBy7QQWEAeZDgBCUAgIAmAPgs6z-gMDDNTVmAFchYggBOGsQWwdkOAwFGAxPXhRKGjoXN09k1PSPTN4YKEEAcy8ACjU4OF5qWgIALhQserhZfI9213cPNQBKRC6oDBgwqBi0TAA6AEkBNKheDBIYGoBtJpyCchl5gHlZACsNraQAfWuewfa5jHmAGWBVuxr7z2GGYYBdYZqBghVSaPxBSFQoy+ACiCgI4GEGF80LR6NB0kwcAAanZgB4iBgPPgKgJKnAALx4QjEEj2LEYCDMXH4wlpEnlKo+VQkaJlVnCAlEzlkinUuIJGpPeakqo1YY4TgANzZIs4QPUqnCkWA0TgqqF7IwNWVl2g7TKYuG7TxRpFcvJY1UDQmUxmBvNUDgBEyduFHMdlWBoJ1mz1MTAZK2hrCGFtasDXPJoyQIK1aijghjiZNKtzHg1o3BEAA1pnozVODRhMIIBqGo2m824ODmGJhEA)

```typescript
import { Type, Static, TSchema } from '@sinclair/typebox'

// ---------------------------------------------------------
// Nominal<T, Brand>
// ---------------------------------------------------------

function Nominal<T extends TSchema, Brand extends string>(
  schema: T,
  brand: Brand
) {
  return Type.Intersect([schema, Type.Object({ __brand: Type.Literal(brand)})])
}

// ---------------------------------------------------------
// Example
// ---------------------------------------------------------

type ValidatedString = Static<typeof ValidatedString>
const ValidatedString = Nominal(Type.String(), 'validated')

function validate(vector: string): ValidatedString {
  return vector as ValidatedString
}

function print(value: ValidatedString) { }


print(validate('validated')) // ok
print('hello')               // fail
```

Would the `Custom Type` example above provide enough of an integration point for the `@coderspirit/nominal` package?

package.json
sinclairzx81: I think for type logic like this, the preference would be to implement the types and inference logic inside TypeBox itself rather than to take on remote dependency.

spec/types/branded-string.ts
sinclairzx81: I've actually been thinking a little bit about what would constitute a nominal type in JSON Schema form. Typically types in nominal type systems are derived from their simple name, so for example the following two types would be seen as distinct in such a system, in JSON schema terms, their names would be implied by their `$id` properties (which is the closest JSON schema has to naming a type)

```typescript
const Vector = Type.Object({
   x: Type.Number(),
   y: Type.Number(),
   z: Type.Number()
}, { $id: 'Vector' }) // $id === brand

const AnotherVector = Type.Object({
   x: Type.Number(),
   y: Type.Number(),
   z: Type.Number()
}, { $id: 'AnotherVector ' }) // $id === brand
```
Unfortunately, both TypeScript and JSON Schema would see these types as one in the same, irrespective of the `$id`, the  types themselves are structurally equivalent. I think for the `Type.BrandedType({...})`, the `Brand<string, 'Email'>`, the `Email` literal would be inferred from the types `$id` property. Such that.

```typescript
const Email = Type.String({ format: 'email', $id: 'Email' })
```

src/typebox.ts
sinclairzx81: TypeBox currently provides no facilities for extending types, instead opting for `Type.Intersect([Base, Derived])` which is inline with the structural interpretation only. Not sure if TypeBox can support a base type of this kind (I think there would need to be proper nominal typing support in TypeScript first, as well as some way to adequately capture the relationship between the `Base` and `Derived` type in JSON schema.

castarco:
Hi 🙂 , still "studying" the code of your comment, but I can advance some comments on specific details in the meantime.

There are some details making `@coderspirit/nominal`'s brand types a bit special:
  - They use symbols for their keys (which are unique and can be declared only once), instead of strings (so using `__brand` as a key does not really work with them).
  - Besides the "brand" marker, there's an extra marker that "saves" the original type (also behind a symbol key), to allow complex type operations (so it becomes possible to combine "brands" and what I call (because I didn't find a better word for it) "properties" (I leave a comment about them as an "annex").
  -  Although it's possible to have a runtime representation for them, I wanted to not make it a requirement (only if we want some kind of reflection). We just need one single cast at validation time, and the type system "lies" to everybody for us, without leaving a trace of their existence during runtime, the types can be wiped out at transpilation time.
  - A minor detail is that I declared the metadata keys as `readonly`.

About why I decided to use symbols for the internal details of `@coderspirit/nominal`, it is because I wanted to ensure that they can't be represented in JSON, so:
  1. they can't be accidentally leaked to the outer world (no need to clean values at runtime).
  2. They can't be crafted from outside by external API users. So there's no risk that our validation layers could be fooled by specially forged requests, at least not because of those nominal types, of course.

I'm not against having an external representation that can be expressed via JSON schema, of course, but I would try to avoid a situation where we have a direct translation between the two worlds, to avoid the previous problems I mentioned.

Regarding this
> This unfortunately results in the output schema being non-valid as JSON schema can't intersect both string and object via allOf.

Completely true, I'm aware of the problem... What I "forgot" is that TypeBox is not in the validation business, and it's precisely there where the casts have to be performed. So the endeavour becomes a bit more complicated, there are more moving pieces than I anticipated. AJV and friends can be quite far more challenging given their size and widespread use.

## Potential solution (for the brand generalization)
Yesterday night (Berlin time) it came to my mind that maybe we could reach a compromise solution. The truth is that there are many "branding" libraries out there, and although I made mine because it is my believe that most of them were not enough, they still deserve consideration, and it would be amazing to have a solution that keeps compatibility with all of them.

Here there's copy a link of the playground (the code got too big to post here):
[playground experiment](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgFQJ5gKYBo4GUYCGMwAxjsriQBYYgHkAywMGUBANuQGocCu2KAJIA7FlADOGEjHIB5AEYArKTLyFiJJmI7l8UYMIDm5AHK8Q81uREtDrOAF84AMygQQcAOQABcQZLsBMBQAPQw6BjyEAAengCwAFCJISFwguDQLAAmcIq84vDO0HAwNHAY0QTg7AIA7hhwtRC87FnCnvDCGBg5pbRwBMKoINANBn5ZDWiYAEIxiaCQsIhwAPqrMwSS0wLrM2zCOU6u7l7eJBCTEmDBzCHC7gYcALTiw1Hs4vFJCSlwzwDAUDgSDQWDwRCAclUgARAhQADWcBmgRISIAsgRDKQAFxwdEYQj7QZZcTQ-6QylU6kgxLkwRwBEPWolGiSEpNAZQBrcrK8Q6DGAAOhFcHkvHgAAoLsISBgwDBeBx2KgAJSsjCoODcsDcySiOBZYDOZysDAG0oGQziIWJcKYfHwhGsHZwAC8cAK+iMcAAPp73hB2HaIsiDllXR6vVa-QGLEG6b9UgyKjA2NINXAAER0RGscRZsVSAj5Br1TzsdjajCljUeQJiEMOglE8OYvMSAA8iTgyK2GB2WB7YZJ5WiLEO4hHh0Hw+Jh3bzqgY4npMdHdnCV7m22EUX9lT5rX+6grv9-MmzgMPXdcAvGCvXSyQ4SAD5bzuB3unQfx0epyeOzDr2AD8KzcgQWQQMIKpwAA2kiBjTlkJ4ALp4vORxwAAZOBGCQdBsEIQATHASGfjsaEYf2roOMBcB4kgEFQTBWoIWRwjIVRyGOIm9oNC2BCYd2W59rumAvtu4Yrv+yGblJJInjJk7rku8liV+mBKYeKmAaG56HA+145B696Pj0L7vh6FGhrhgmYSe4idjZEnITgDk-lA7k0d+HavomfzIFQwBTk0LQ5JYAzZryRnCMwwDQYWEDOHA5yXPmNz6DA9yPMIHCeFODwgE8Vb8WSCT8XAJi5RwwkuQImHKWumE7FZ+KEkJ4bOT5rmYTg-HJWsGzhv1ESDXsPUYP5PyBcFU7yNJlUhXAAQQJIy4wBAcBUBALILaOZVcg0kziCQ+iWBFWq+P4gTBAAXtEAAcACMTZTBEczRHVk3udJOnNeGrW3vZXX1b9JI4J46z7YcnjTeSNKI0j4LktVxV5ewnbIODhzTX8yME4TAXJit0EFEET6ZpVlgkCW7KCJ4HgPPA4i8NyAxRBKcCGBAMYGJtmZsPI8jMNtQYNINRUlSUERTkUy4XNUFRwBASgqOVlUUIsNS3hQaYxv6yBmBY9iGzYGB2MuhtaKwHBYzw7D8O+cB-HM4uDFOLEdIaGDrQAbmMMDE9mADqDRdD0hYEHAIzs3YXRsFWAcSAlnGDZaU7QQIWxpKT4VwMzAxZDkuoQJgUDkqzYBLPAAt9BplGeQxkqJM4-LSKncAAOLmrb7DAHdPRoyVIm9pQNB0K6-1TlrSuSTx0+evrRgvsOs0NLHDSK2A-e09kMsOmA8JVIS+Zzm2TeL3pElzpN2l-rpnlnnehnmc+iSvi3omnRPBB4uPtACDqRhlkDCI1Ey9nXuyGAnJxA7XzlQAgAcvDQ3DJ4AYhwX6XmMgMT2D4SzsBgOIeefxxTwAoAAYTIlOfkBB5A6wFveCQhBMGlCIGRFK9RDTQW9kfcQU5o5+z4NgNeqQ+icXrodOBzRWhFg4qaKAvIXBuA8PXS8BD4BHzYCAU+EhRFwElMaDU7Nag52jsVcQdAYDUBwKLMKsjIocFMagKckB+HAHoQ0AWe9qBkXgA2VgqphwgJPGAxSnl57yDvp5ai4kMAnhfKqPEyBzYSBUJ2OCw4AGTwiPPZAChlDSE7ExfCLEiKIU4h5Ds6EUA20Ts5cM74nC4UlPVe+q4AJPwiPRMCJSCKsXghUhuvklw1OyUAiIcFPAABJyYaE8KhXioleyMQcKqaaqF3wIH0WHbaSCvFbUihUHepBRZ0OaLXMo3JFRQAkaGGmdMyxb0GCrAZTJdoan0WmfgB8GiWGCpgqpS5lGnFqDQExzzKx+JwF0ZOIKPDjGAJMYcw5rls04jsIUqTJDSElJk5ZnpqCAPnr2TFBSVBf17FSxA9FqVwRCZ5GpmK6kcElCA1Uq8CXUr+FmPhfgfRQVqMIQs9cDBFA5hcsU-Zfnc0JDQwwsFdSEnCC4LYMhaVUr+AzJmEAWZswaDtFkAs-DVGNFqDOHFMwkDZtyA0MoWDjkaMweBQcuWatSCMI0zhUAxkNSgD6MRGjQARDaDVvYRRCjZdEjscBen0ujcCnO18MA1J-oAqZsz1CkAWQ4VZHKNVOBzn0spbEhlAtYEyllmNMJNJwgYtpl8H7Hi6ZgHpeF+nlMtQ26p-8iU5MwBmuZ2bFl0QJXml2qQgrLVpgUahAx2DOKnBcRRKgcBkJQLgKhUFfYF11UWfuGBkH5HokE0SqFT2jtbu3Yg0Eqo1UxsgJqM9cDawan9JtU5oxGE-sONNdBkmRPDGEw4iR1TbNElqsWDo+jcgKp6dwXj0Axn1miVk7CLj52VB8-IJQtpIN5r0Oa+jUxsBcNezutNKxClzkaNo3tHw5FFqgQkOB+Ty0VHlFgKpbSiTRbc7uvdE4DyHveyUf76BihGig1YIDPCsdfsZC9wdCYqcRuSAAopUJWCNVO6chImGUs6HZIqID0PQMYPTDwxpKTF5mjCSg5V4IR-csimayJ4U9lVjOueyHZwwt58BEFIJ2AaKVvNub8-DBIbdZQ3s4s5kzLBJQB2kNAPEX7DBJLgOF3zy9-Pgd7Hx+LKhig5xy2ZvLiRL3RfI7e3U-NkvCLxOVrIfmwNLMTPV0QjWXOmclJ4BLPmegefVH8CACJEhdZgP1mglYIAeepYtpbfxnBBGDDNVIemttgg002m8mmqg7wwDp7bp2oQ-Bix3W9uBAyfCs3bR9i9Z5HZxjkReGWf3fz7X-FAgGSTAayKBmlvHCTooEwnDgwmsj3fYGJ77a6pN7ERxsSaSmNtpBgHBjgIxZ1gvYQyUxogLJii5mC4ADCyjqboOT0mwhZ07VaOVP4SFmBTkqvXZd3IMxiq2vLYZrlRaSFoGzxB8BmDkjsEQuA7A1rwGgnKRoPJ8K+PrlTtbvycClyifQrUSE+jkhSUTtJGYCjymo9VPV3J+rwMMFQHjkHDD8H4X4z08DWjtHgJFbeu83O4ZcMAaI8L-WzEDRMERSYSfwAZLRj3jRBTVlqFlBoWi5cpVZxxMQGNpcQGxCQFwxQM7kgqIdmoPHDPwDVzTj0N34x3dE7ZvLDnIaV-YB5t6cAW8BazSQELY0Uot6i5duLcBBumZb5KQB5P0t5ay53gr1YbmcUn1WHOLequJiH53Kb4-l94hb+16r2-qew9HywcfnhZsy+8E0KArQRuTe9DAc-l+IDX+gHf09QA)

Basically I define the concept of "MetaBrand", to allow people using different branding strategies. When it comes to the "base type" I wanted to preserve, I only had partial success, but I hope the example can be useful to continue the discussion.

## Annex
### About the weird "properties" I introduced in Nominal
These "properties" are a bit special because we can combine many of them at the same time (as opposed to what happens with brands and flavors), so we can "attach" many properties to a same value. For example, we could "attach" the properties `Integer`, `Positive`,`Odd`, and `Prime` to the literal value `7`. And this is made in a way that:
  1. Some properties can be defined to be the opposite of other properties (like even vs odd).
  2. we can define function signatures that "know" how their transformations will affect the properties of the returned values (without having to write all the combinations as function overloads).

sinclairzx81:
@castarco Hi, apologies for the delay in reply, and thanks for the follow on response.

While I do actually like the idea of nominal types in TypeScript, as of current, adding them to TypeBox is out of scope essentially because TypeScript doesn't implement a nominal type system (with nominal types only partially expressible with primitive types via user defined workarounds). I think the expectation would be that nominal types should be supported for all types (classes, objects, functions, etc), but this currently isn't possible in TypeScript (as of writing), and I'm pretty reluctant on introducing partial support through user defined patterns for primitive types only.

The other problem is that JSON schema is largely structural (it's not nominal either), and in this regards; it's not clear if TypeBox should implement nominal types even if TypeScript were to provide a `nominal` specifier as there wouldn't be a way to differentiate between two schemas of the same structure (not without adding discriminators to that structure). JSON Schema would see two exact schemas as equivalent in the same way TypeScript currently sees two exact interfaces as structurally equivalent. In this regard, introducing nominal type support may result in a mismatch problem where type equivalence would diverge between TS and JSON schema. Such mismatches are less of a problem if the expectation is that type comparisons are always structural.

As mentioned, it should still be possible to implement a nominal type externally without baking a nominal type into TypeBox. In both TypeScript and TypeBox cases, the nominal type can be constructed via type intersection. I feel implementing things this way is generally better as it keeps TypeBox in tune with the capabilities of TypeScript (that is to say its better for TypeBox not to introduce constructs that are not naturally expressed in the language). However allowing for user defined nominal types to be  composed via type composition (as it would be done in TypeScript) seems a reasonable option if nominal primitives are required.

Lastly, TypeScript does actually have opt-in nominal typing on the roadmap, see [here](https://github.com/microsoft/TypeScript/wiki/Roadmap) and [here](https://github.com/Microsoft/TypeScript/issues/202). I think my preference at this stage is to wait and see what Microsoft comes up with first before considering nominal type support in TypeBox (this to avoid introducing functionality now that may be at odds with a language supported feature in future)

Thanks for the PR though, It is interesting to think about how nominal types could work in the library, and possible food for thought if Microsoft do provide some form of opt in nominal type system in future. I'll leave the PR open for a few days and give it a little bit more thought. Interested to know how you got on with the `TNominal<T>` example above.

Many Thanks!

castarco:
Its time has not arrived yet 😅 .

unao:
From my experience having nominal types just for primitives is completely enough but highly desirable, so I am extremely glad I have found this issue. 

https://github.com/sinclairzx81/typebox/pull/145#issuecomment-1019374031

@sinclairzx81 if still possible, please, replace in your snippet `'$static'` with `'static'` when accessing `TSchema`.

btw. for me using `Symbol` instead of `string` is preferable: it makes the `__brand` inaccessible.

```typescript
const __brand = Symbol('__brand')
type __brand = typeof __brand

export interface TNominal<T extends TSchema, Brand extends string> extends TSchema {
  static: T['static'] & { [k in __brand]: Brand }
}
```

castarco:
@unao far from perfect, but some time ago I introduced https://www.npmjs.com/package/@coderspirit/nominal-typebox , it has to be combined with https://www.npmjs.com/package/@coderspirit/nominal (so it's not completely "lib-agnostic"), but it works.

Issue #107, opened 2021-09-17
Better support for `enum: T[]`


jonahsnider:
A limitation in OpenAPI v3 means `const: T` doesn't work, and instead `enum : [T]` is recommended as an alternative.

The documentation only lists one way to get something similar to that:

```js
const T = Type.KeyOf(
  Type.Object({
    x: Type.Number(),
    y: Type.Number()
  })
);
```

Could there be better support for creating enums in the resulting JSON? Maybe by adding an overload to `Type.Enum()` to accept an array of literal values? Or allowing several values to `Type.Literal()`?

sinclairzx81:
@jonahsnider Hi, the current approach for extending TypeBox to support specialized OpenAPI types is to create a function that returns the schema you need, and reinterpret as a TypeBox `TSchema` type. Consider the following example that returns a OpenAPI `enum` type (as per issue), but is reinterpreted as a TypeBox `TUnion<[...TLiteral]>`  type for static inference.

```typescript
import { Static, TLiteral, TUnion } from '@sinclair/typebox'

type IntoUnion<T> = {[K in keyof T]: T[K] extends string ? TLiteral<T[K]>: never }

export function CustomEnum<T extends string[]>(values: [...T]): TUnion<IntoUnion<T>> {
    return { enum: values } as any // Note: it's possible to return OpenApi
                                   // compatible schemas and reinterpret
                                   // as TypeBox TSchema types. This type
                                   // is not strictly JSON schema, but so
                                   // long as it validates with the same
                                   // behavior (as per OpenAPI spec), we deem
                                   // this safe.
}

const T = CustomEnum(['Foo', 'Bar', 'Baz'])

type T = Static<typeof T> // type T = Foo | Bar | Baz
```

Also, take a look at issue #105 which goes into a bit more detail on mapping types this way, and includes an example for implementing a `enum` type also. 

Hope that helps
S

sinclairzx81:
@jonahsnider Going to close off this issue. Feel free to reply here if you have any problems or general questions on extending TypeBox in this way. 

Cheers
S

jonahsnider:
Thanks for the snippet. Seeing as this is a common use case for the library, would you consider making it an official feature? Or creating a `typebox-extra` package?

sinclairzx81:
[...]

Issue #59, opened 2021-03-18
Question: Enum and Unions


btd:
Hi, thanks for awesome project.

I have a question about union types and enums. We do not use ts enums because they are not supported in babel typescript. But for example such simple type:
```ts
type M = 'single' | 'married'
```

Could be expressed as:
```js
{
type: 'string',
enum: ['single', 'married']
}
```

So that if i would write it like this with typebox:
```ts
const M = Type.Union([ Type.Literal('single'), Type.Literal('married') ])

/*
{
  kind: Symbol(UnionKind),
  anyOf: [
    { kind: Symbol(LiteralKind), type: 'string', enum: [Array] },
    { kind: Symbol(LiteralKind), type: 'string', enum: [Array] }
  ]
}
*/
```

so my suggestion is to allow to put union of literal values of the same type to be encoded as enum. WDYT?

btd:
Probably it makes more sense in this case to define new type, like a mix of Literal and Union:
```ts
export type TLiteralUnion     <T extends TValue>   = { kind: typeof LiteralUnionKind, enum: T[], type: 'string' | 'number' | 'boolean' }

public LiteralUnion<T extends TValue>(items: T[], options: CustomOptions = {}): TLiteralUnion<T> {
        // there missing check items.length > 0
        const type = reflect(items[0])
        // there missing check type not unknown
        return { ...options, kind: LiteralUnionKind, enum: items, type  }
    }
```

sinclairzx81:
@btd Hi, thanks for the suggestion :)

I do actually like the JSON schema `enum` representation. However at the moment, I'm planning on keeping on with the current `anyOf` implementation of as it generalizes well across all types. 

I think an implementation of `enum` would likely sit under `Type.Union()` as a optimization of the current schemas (as opposed to adding a new Type to support this usecase). Also, I am currently looking at what is involved in providing a `Type.KeyOf(...)` implementation and trying to avoid complications with respect to union handling as `keyof T` would map to an `TUnion`. For example.

```typescript
const T = Type.Object({
   x: Type.Number(),
   y: Type.Number()
})

const K = Type.KeyOf(T) // TUnion<[TLiteral<'x'>, TLiteral<'y'>]>

type K = Static<typeof K> // 'x' | 'y'
```
[...]

Issue #155, opened 2022-02-22
Cannot Reference schemas created using the StringUnion function from the "OpenAPI" section of the Readme


jkennedy1980:
Hello, Thanks for the great library.

I've found that you cannot use `Type.Ref` for schemas created using the the `StringUnion` method from the 'Open API' section of the documentation. The reason is that `Type.Store()` never gets called since the schema is directly returned. Here's a single file the demonstrates the problem:

```
import { TLiteral, TUnion, Type } from '@sinclair/typebox'

enum TestStringEnum {
    one = 'one',
    two = 'two',
    three = 'three'
}

type IntoStringUnion<T> = {[K in keyof T]: T[K] extends string ? TLiteral<T[K]>: never }

function StringUnion<T extends string[]>(values: [...T], $id?: string): TUnion<IntoStringUnion<T>> {
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    const schema = { enum: values, type: 'string', $id } as any
    return schema
}

const stringUnionSchema = StringUnion(Object.values(TestStringEnum), 'TestStringEnum')
console.log('stringUnionSchema: ', JSON.stringify(stringUnionSchema, null, '\t'))

const stringUnionSchemaRef = Type.Ref(stringUnionSchema)
console.log('stringUnionSchemaRef: ', stringUnionSchemaRef)
```

**The code above with throw an 'Unable to locate schema with $id 'TestStringEnum' when the `Type.Ref()` is called.** 

I was able to work around this issue by extending `TypeBuilder` and calling `TypeBuilder.Store()`:

```
import { CustomOptions, TLiteral, TUnion, TypeBuilder } from '@sinclair/typebox'

enum TestStringEnum {
    one = 'one',
    two = 'two',
    three = 'three'
}

type IntoStringUnion<T> = {[K in keyof T]: T[K] extends string ? TLiteral<T[K]>: never }

class CustomTypeBuilder extends TypeBuilder {

    StringUnion<T extends string[]>(values: [...T], options: CustomOptions = {}): TUnion<IntoStringUnion<T>> {
        const schema = { ...options, enum: values, type: 'string' } as unknown as TUnion<IntoStringUnion<T>>
        this.Store(schema)
        return schema
    }

}

 const Schema = new CustomTypeBuilder()

const stringUnionSchema = Schema.StringUnion(Object.values(TestStringEnum), { $id: 'TestStringEnum' })
[...]

PR #172, opened 2022-05-03
Add string enum

noamokman:
Hey, I want to add a new type called `StringEnum`.
As previously discussed in #59 and #107 and some others,
The reasoning is if someone wants to use an enum for string values only, then the schema will be:
```ts
const T = { 
 anyOf: [{ 
   type: 'string',
  const: 'a'
 }, {
  type: 'string',
  const: 'b'
  }]
}  
```
but for open api and server validation errors, it makes more sense to have:
```ts
const T = { 
   type: 'string',
  enum: ['a', 'b']
}  
```
Also, using this we can have an enum as the type, changing to union strings from the section in the readme changes the behavior.
With this implementation we can still use enum as the type and get the desired schema


sinclairzx81:
@noamokman Hi, thanks for the PR.

Unfortunately, I still don't think I can't merge this one through. The preference for TypeBox moving forward is to have one representation of union across all types, and future versions of TypeBox will be unifying around the current `anyOf` representation (including actual TS enums) in future releases. The reason for this is there are some compositional considerations and complexities around union composition that become difficult to reconcile if TypeBox has multiple underlying ways to expression Union. Also there is work being done upstream to make the composition experience better that is only feasible under a single representation. 

## Current Functionality

As per https://github.com/sinclairzx81/typebox/issues/107, it is possible today to extend TypeBox with custom schemas for specifically for OpenAPI if you need. The following creates a custom `Nullable` and `StringEnum` type by extending the TypeBuilder.

TypeScript Link [Here](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAKjgQwM5wCoE8wFMBCEAHnAGZQQhwDkAAqsAHYDGANssFAPQw64BGxagFgAULiKRYiMhSp0GLdpx59BREaLG88cAJKMYEAMowoTAOYAZYDFxRkrAKqNgERgB4MAPjgBeOABvAG0AaTgmOABrXCwIUkwAXQAuTDDEuAk7RgATdFQzSzgAfkw+QiIAOgwbOwdWL3TvVMZcADd7OABfMTEJKXg2NHQAeTxGAEEwYGw8fABXYFYczqzcXPRZgmJq8sXlzsDe0ThTuDB5-lZgZjgAOXnWdivcL0yibLyyuZ2MY2YABa4EDIbwAClQgOByFSGAAlKlvtsqhgXG5PMEMAAaJEVaoPJ6JXxHE5nMlQXAweZQRhBOCVBmQoEgnGMR7PVi4VJmea4boodDIRhYMRknqk06is4XK43OCmcyMCwAUTZIDeaw2cAKioswSJYLajl5qFSwQZ1USCNxvzR7g8BiMCsstXsjjtnh8xKlZNOFKpNLp63mIFSRtYJv5aBQwrgPu6YnFYi4XDgAFoM5ms9mc7m8znk6nlURkOBOYX0-mq9Xq8d+tBBu4Ckj-HBWgB3OBjdZTGZ7JYrKBwMFw47MJvwDCtraVEb8ABWuGYMDBJNOMKRlQJHNwYJnzqVI7hWKl-FhfEqB5VarBwWoE2oOOo+EfNAAwtQrYnR1pRDo+VOARbHipjIDANweP+8SYN4YhAA)

```typescript
import * as TypeBox from '@sinclair/typebox'
export * from '@sinclair/typebox'

type IntoStringLiteralUnion<T> = {[K in keyof T]: T[K] extends string ? TypeBox.TLiteral<T[K]>: never }

export class OpenApiTypeBuilder extends TypeBox.TypeBuilder {

    public Nullable<T extends TypeBox.TSchema>(schema: T):  TypeBox.TUnion<[T, TypeBox.TNull]> {
        return { ...schema, nullable: true } as any
    }
    
    public StringEnum<T extends string[]>(values: [...T]): TypeBox.TUnion<IntoStringLiteralUnion<T>> {
        return { enum: values } as any 
    }
}

// ------------------------------------------
// Example
// ------------------------------------------

export const Type = new OpenApiTypeBuilder ()

const T = Type.Object({
  a: Type.Nullable(Type.String()),
  b: Type.StringEnum(['A', 'B', 'C'])
})

type T = TypeBox.Static<typeof T>
```
## Future Functionality

In the next release (planned mid June), TypeBox will provide a new `Type.Unsafe(...)` API to make creating custom schemas a lot simpler. Note that this new API is being trailed and refined in an auxiliary project which you can find [here](https://github.com/sinclairzx81/sidewinder), with The anticipated usage for this API is as follows.

TypeScript Link [Here](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgFQJ5gKYBo4GUYCGMwAxjsriQBYYgFwC+cAZlBCHAOQACAzsABMMAd2AA7IVAD0MdBk4BYAFDLmAVzEliEMXAByagDaGCAI0MYAPMjgYAHjAwTeKSjToA+ABS9qtAgBcKACUiMpwcFAYMGpQumiYAHQAqmK8BMxW+ESk1h5wAD5wYkaG3kglxmYWQTBQathwic2+7vQMwcoMyqoaWsA6eHXiAOYAoiUg1rYOTgIuvMNiIwDaALreAG4Ehg28QSvNichroQjhkdGx8XIpaRlWyCuTphhQG15ITmogQdu7GBcHS6PSUUikcAAtNCYbC4fCEYjEcpwXAxnYCOALCiIUi8fiCdDQSQdIsUHAALwoW4AeVMACsMFpPhdAvpStUMF4EhhEvgoKMvMFglgLqYgvzRhMfl4VpwAIKcHCcABCSq4AGFOKcup0VEpZJhyVTssQSJZDRgIMwUB5QUA)

```typescript
import { Type, Static, TSchema } from '@sidewinder/type'

function Nullable<T extends TSchema>(schema: T) {
  return Type.Unsafe<Static<T> | null>({ nullable: true, ...schema })
}

function StringEnum<T extends string[]>(values: [...T]) {
  return Type.Unsafe<T[number]>({ enum: values })
}

// --------------------------------------------
// Example
// --------------------------------------------

const T = Type.Object({
  a: Nullable(Type.String()),
  b: StringEnum(['A', 'B', 'C'])
})

type T = Static<typeof T>
```
Would either of the above solutions provide a satisfactory way to create schemas you need for OpenAPI?

sinclairzx81:
@noamokman Thanks again for the PR, it is very appreciated. But as mentioned, I can't accept due to the varying JSON schema representations for Union. As this would introduce multiple ways to express Union, it actually implicates software down-stream that may need to reflect on these schema representations. There is already a fair amount of sophistication involved in interpreting Union in some of the code I've written in this area, so the preference is to keep things as simple as possible (where the `anyOf` representation is generic to apply to all things expressible in TypeScript). This reasoning may change in future, but not in the foreseeable future.

Moving forward, the recommendation will be to leverage the new `Type.Unsafe(...)` API for custom schemas (either normalized (i.e. string enum),  or aligning to non-standard schemas (such as OpenAPI 3.1< handling of nullable). This function will allow for the unrestricted creation of custom types, and reduce the overhead in creating types without necessarily having to involve other parts of the TypeBox infrastructure. 

Again, thanks for the contribution, these are always welcome and always worth discussing.
Regards
S

noamokman:
Looks like the `Type.Unsafe` functionality can also help with #155,
The thing is that I use fastify + fastify-swagger + typebox in order to support openapi,
I later use the openapi to generate a typescript client.
In order to both use an enum in the server and in the client I have to use the provided solution.
I don't think that this is a unique solution and we would have this functionality in every microservice.
If the `Type.Unsafe` lands in typebox I would probably warp the string enum in a separate package to offer it to people who need it as well.
Do you need help with copying the implementation here? it looks pretty straight forward.

sinclairzx81:
@noamokman Thanks, but that's ok. The implementation for `Type.Unsafe(...)` is due to be released around mid June with some other enhancements (such as support for recursive schema inference and a few other things). The schedule for this release is somewhat tied to the TS 4.7 release, as well as some availability in my schedule to roll these updates through. 

You can essentially use the `Type.Unsafe(...)` API today by taking the `@sidewinder/type` package as a dependency over the `@sinclair/typebox` package (as the `@sidewinder/type` package is being used as a testing ground for the next version of TypeBox prior to formal release).

### OpenAPI

It's worth noting that TypeBox isn't specifically written to support OpenAPI out of the box. The reason for this is due to fact that OpenAPI versions previous to 3.1 were extending the JSON schema specification to the point of being non-standard JSON schema. This relates to OpenAPI's handling of `nullable` (as well as a few other things). TypeBox intentionally makes these disparities apparent to the user, as you've seen.

From TypeBox's standpoint, rather than implementing the requirements of OpenAPI (which would require TypeBox to support multiple schema specifications), TypeBox opts to generate schemas that conform to the JSON Schema specification only. It does however provide various escape hatches (such as `Type.Unsafe(...)` and other approaches above) for OpenAPI users who need specific representations not created by TypeBox by default.

The `StringEnum` representation fits inside the category of `custom` schemas. Both in terms of how the representation is defined (TypeBox will preference the single representation over the ambiguity of multiple representations), as well as viewing such custom schemas as an auxiliary requirement (in this case OpenAPI)

I feel things may become easier once OpenAPI 3.1 fully adopts the JSON Schema `2020-12` specification.

sinclairzx81:
> If the Type.Unsafe lands in typebox I would probably warp the string enum in a separate package to offer it to people who need it as well.

This is actually encouraged. For example consider https://github.com/sinclairzx81/sidewinder/blob/master/libs/contract/type.ts#L92-L100 which extends the default TypeBuilder to provide a custom `Type.Contract(...)` function to which is used to specify WebService IDL. This package is shared between both Client and Server. It's expected that a similar approach could be used to create a OpenAPI IDL builder, which would mean  users could take some specific `OpenAPITypeBuilder` package and not the defacto TypeBuilder provided by TypeBox (which only creates a finite set of JSON Schema representations)

Note: The OpenAPITypeBuilder has been considered before, refer to https://github.com/sinclairzx81/typebox/issues/107#issuecomment-923452287 for a high level example of what that might look like.

SimonPiscitelli:
normal jsonschema:
```ts
{
 type: "string",
 enum: ["foo", "bar"]
}
```

can't this be

```ts
Type.String({enum: ["foo", "bar"]})
```

or
```ts
Type.Enum(["foo", "bar"])
```

PR #178, opened 2022-06-11
Add `isTSchema()` type guard

jonahsnider:
Adds a type guard to check if a value is a `TSchema`. Additional type guards could be created for each specific schema kind (ex. `isArraySchemaKind()`)

### DIFF
--- a/src/typebox.ts
+++ b/src/typebox.ts
@@ -63,6 +63,19 @@ export const AnyKind       = Symbol('AnyKind')
 export const RefKind       = Symbol('RefKind')
 
 // --------------------------------------------------------------------------
+// Type Guards
+// --------------------------------------------------------------------------
+const ALL_SCHEMA_KINDS: ReadonlySet<symbol> = new Set([
+  NamespaceKind, KeyOfKind,   IntersectKind, UnionKind,   TupleKind,  ObjectKind,
+  RecordKind,    ArrayKind,   EnumKind,      LiteralKind, StringKind, NumberKind,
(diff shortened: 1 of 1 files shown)


sinclairzx81:
@jonahsnider Hi, thanks for this. And apologies for the delay in response (have been rather flat out recently)

I think something like this could be viable, but perhaps not in the current version of TypeBox. The `kind` property is flagged for a change in the `next` version of TypeBox, and the `StringKind` and `NumberKind` symbols are to be removed. 

This change is outlined below.

```typescript
// -------------------------------------
// current
// -------------------------------------
const StringKind = Symbol('String')

const T = { kind: StringKind }

// -------------------------------------
// next
// -------------------------------------

const Kind = Symbol('Kind')

const T = { [Kind]: 'String' }
```
Would it be possible to retarget this PR to point to the `next` branch and with the following amendments?

- Create `src/guards/index.ts` module.
- Move the `isTSSchema()` logic into a `namespace` called `TypeGuards` inside `src/guards/index.ts`
- Reevaluate the `Set<symbol>` for `Set<string>` logic.

There has been some considerations recently to have TypeBox provide some form of opt in runtime assertion logic, and type guards may fit into this picture. However, I think for this feature to roll out, there would be a necessity to provide guards for all the primitive types (not just `TSchema`). There may be benefit in formalizing all the schemas produced by TypeBox by encoding that in the guards as exhausted checks. So could be worth exploring. 

Let me know your thoughts
S

PR #181, opened 2022-06-25
next: Namespace `symbol`s to avoid collisions

jonahsnider:
Because `Kind` and `Modifier` are common strings, it is reasonable that someone could accidentally get the same symbol for them via `Symbol.for()`.

This PR "namespaces" them by using a `typebox.` prefix, to avoid accidental collisions.


sinclairzx81:
Hey thanks for this :) Will merge through.

PR #182, opened 2022-06-25
next: Type guards

jonahsnider:
Closes #178 

Creates a `Guards` namespace with type guards for all the types of schemas as well as `TSchema`s.

The package entrypoint was changed from `src/typebox.ts` to `src/index.ts`. The guards module imports type from `src/typebox.ts` and I didn't want to have circular imports/exports if `src/typebox.ts` was used to export the guards module.

I tried updating the tests but couldn't understand how the build system for them works.


src/index.ts
```diff
@@ -0,0 +1,30 @@
...
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+
+---------------------------------------------------------------------------*/
+
+export * from './typebox';
+export * as Guards from './guards/index';
```
sinclairzx81, line 30: Ah, let's omit the `export` for the guards here.

Note: the `compiler`, `guards` and `value` modules are new and intended to be imported with the following.

```typescript
import { TypeCompiler } from '@sinclair/typebox/compiler'
import { TypeGuard } from '@sinclar/typebox/guards'
import { Value }  from '@sinclar/typebox/value' 
```
The reason for this primarily stems from having the ability to tree shaking where users can optionally pull this additional infrastructure if they need it. This mostly relates to the compiler module (where users would typically be using Ajv, and thus should be optionally bundled).

The general rule moving forward is that if there is any runtime functionality (or anything that results in additional code that isn't erased through compilation), then it needs to be treated optionally.

Would be open to additional approaches if you have any ideas tho.

src/tsconfig.json
```diff
@@ -12,5 +12,6 @@
     "compiler/index.ts",
     "value/index.ts",
```
sinclairzx81, line 13: Lets include `guards/index.ts` here. This enables TypeScript to emit the code for the guards without importing through `src/index.ts`.

tsconfig.json
```diff
@@ -13,7 +13,7 @@
         "src/value/index.ts"
       ],
       "@sinclair/typebox": [
-        "src/typebox.ts"
+        "src/index.ts"
```
sinclairzx81, line 16: Let's keep this as `src/typebox.ts` and remove `src/index.ts`. Deno users currently import using this name as a path so will keep it as is for the time being (although it is a little non-standard)

src/guards/index.ts
```diff
@@ -0,0 +1,86 @@
...
+export const isTRecord = typeGuardFactory<TRecord>('Record')
+export const isTSelf = typeGuardFactory<TSelf>('Self')
+export const isTRef = typeGuardFactory<TRef>('Ref')
+export const isTString = typeGuardFactory<TString>('String')
+export const isTTuple = typeGuardFactory<TTuple>('Tuple')
+export const isTUndefined = typeGuardFactory<TUndefined>('Undefined')
+export const isTUnion = typeGuardFactory<TUnion>('Union')
+export const isTUint8Array = typeGuardFactory<TUint8Array>('Uint8Array')
+export const isTUnknown = typeGuardFactory<TUnknown>('Unknown')
+export const isTVoid = typeGuardFactory<TVoid>('Void')
```
sinclairzx81, line 86: I think this is maybe ok for now. But would be ideal if each guard checked the expected schema structure for each type. So for example.

```typescript
// This is the definition for a String
export interface TString extends TSchema, StringOptions<string> {
  [Kind]: 'String'
  static: string
  type: 'string'
}

// So the guard should check all the minimum required properties for it to be a `string`.
function isTString(schema: any) {
    return schema[Kind] === 'String' && schema.type === 'string'
}
```
Having this would essentially formalize the schemas generated by TypeBox.

sinclairzx81:
Hey thanks for this. See reviews above for a couple of small updates. Can go with the current type guards (checking on the `Kind` only for now), but may spend some time looking at checking the full structure of the schema in later updates.

sinclairzx81:
@jonahsnider Sorry mate. Just noticed this PR was automatically closed after the `next` branch was deleted. I've actually rolled through TypeBox `0.24.0` today (which was the `next` branch) over to `npm` (https://www.npmjs.com/package/@sinclair/typebox). Still interested in exploring these type guards as they would formalize the existing schema representations, as well as be very useful in type reflection use cases.

Happy to collaborate on this work. Feel free to resubmit if you're still interested.

Kind Regards
S

src/guards/index.ts
jonahsnider: IMO this is doesn't provide any real value for users while introducing more complexity & increasing the performance cost of these functions. Checking for the `Kind` symbol should be sufficient, users creating their own schema objects with a `Kind` symbol should be expected to properly conform to the interface defined in the types.

jonahsnider:
Sorry for the delay, I was OOO for a few days. I've recreated this PR with the updated version of the codebase at #191.

PR #250, opened 2022-09-24
Pick/Omit: delete the `required` property if empty

dwickern:
If there are no required properties as a result of `Type.Pick` or `Type.Omit`, produce an object without `required` instead of `required: []`.

```ts
const O = Type.Object({
  x: Type.Optional(Type.Number()),
  y: Type.Number(),
})
const T = Type.Pick(O, ['x'])

// before:
const T = {
  type: 'object',
  properties: {
    x: { type: 'number' }
  },
  required: [],
}

// after:
const T = {
  type: 'object',
  properties: {
    x: { type: 'number' }
  },
}
```
This is relevant to OpenAPI 3.0.3 which refers to [an older JSON Schema Specification](https://datatracker.ietf.org/doc/html/draft-wright-json-schema-validation-00#section-5.15):
> required
> The value of this keyword MUST be an array.  This array MUST have at least one element.

It's also consistent with `Type.Object`, which only adds `required` if non-empty.

### DIFF
--- a/src/typebox.ts
+++ b/src/typebox.ts
@@ -743,7 +743,10 @@ export class TypeBuilder {
   public Omit(schema: any, keys: any, options: ObjectOptions = {}) {
     const select: string[] = keys[Kind] === 'Union' ? keys.anyOf.map((schema: TLiteral) => schema.const) : keys
     const next = { ...this.Clone(schema), ...options, [Hint]: 'Omit' }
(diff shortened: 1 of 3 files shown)


sinclairzx81:
@dwickern Hey, this is an excellent find, thank you!

Will merge this one through and publish shortly.

Cheers!
S

sinclairzx81:
@dwickern Updates published on `0.24.43` :)

Thanks again!
S

Issue #244, opened 2022-09-14
Create schemas from type predicates


BenWoodworth:
Hey! Awesome work with this project and it's been massively helpful :)

Could it be possible to define a type schema from a [TypeScript type predicate](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards)?

Something like this is what I have in mind, I'm curious if there's a way to achieve this now or what you think about adding it to TypeBox (I'd be happy to create a PR).

```typescript
// type predicate (https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards)
export function isMyType(value: unknown): value is MyType {
    return typeof value === "number" && Number.isInteger(value) && value >= 0
    // ...or some other custom type-checking logic
}

// tagged intersection (https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes-func.html#type-aliases)
export type MyType = number & { __myType: never }

// defining a schema based off the predicate (the type is inferred to be TPredicate<MyType>)
const MyTypeSchema = Type.Predicate(isMyType)
```

And potential signatures:
```typescript
export interface TPredicate<T> extends TSchema {
    [Kind]: 'Predicate'
    static: T
    // maybe some others...?
}

// Type.Predicate(...) schema function
export function Predicate<T>(
    predicate: (value: unknown) => value is T,
    options: SchemaOptions = {},
): TPredicate<T> {
    // ...
}
```

sinclairzx81:
@BenWoodworth Hi! Glad you're finding the library useful :) And thanks for the suggestion!

I do actually think TypeBox needs something like this for user defined schemas that do not fall under the standard subset, however the caveat would be the predicate schema would need to be serializable. Unfortunately, encoding runtime logic inside `Type.Predicate()` would result in that logic being lost on serialization (even though encoding runtime logic inside the schema would be super convenient from a compiler and value check standpoint)

Currently, the design for custom schemas is to use Type.Unsafe() with Value.Check() and TypeCheck.Check() acting as type predicates. As of 0.24.x, the following works from a inference type perspective, just not from a functional runtime perspective (as custom schemas are yet to be implemented)

[...]

PR #259, opened 2022-10-13
feat: validation for NaN in number schema

g4rcez:
A feature to accept or non `NaN` in number schema

### DIFF
--- a/src/compiler/compiler.ts
+++ b/src/compiler/compiler.ts
@@ -138,6 +138,7 @@ export namespace TypeCompiler {
 
   function* Integer(schema: Types.TNumeric, value: string): IterableIterator<string> {
     yield `(typeof ${value} === 'number' && Number.isInteger(${value}))`
+    if (schema.acceptNaN === false) yield `(typeof ${value} === 'number' && !Number.isNaN(${value}))`
     if (schema.multipleOf !== undefined) yield `(${value} % ${schema.multipleOf} === 0)`
     if (schema.exclusiveMinimum !== undefined) yield `(${value} > ${schema.exclusiveMinimum})`
     if (schema.exclusiveMaximum !== undefined) yield `(${value} < ${schema.exclusiveMaximum})`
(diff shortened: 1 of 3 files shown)


sinclairzx81:
@g4rcez Hi, thanks for this PR.

Unfortunately, I can't really merge this through as JSON Schema does not support `NaN` values, and by definition `NaN` is "not a number" so by including such a option, it would permit any permissive value `isNaN` would evaluate to `true`. For example.

```typescript
isNaN(NaN);       // true
isNaN(undefined); // true
isNaN({});        // true

isNaN(true);      // false
isNaN(null);      // false
isNaN(37);        // false

// strings
isNaN('37');      // false: "37" is converted to the number 37 which is not NaN
isNaN('37.37');   // false: "37.37" is converted to the number 37.37 which is not NaN
isNaN("37,5");    // true
isNaN('123ABC');  // true:  parseInt("123ABC") is 123 but Number("123ABC") is NaN
isNaN('');        // false: the empty string is converted to 0 which is not NaN
isNaN(' ');       // false: a string with spaces is converted to 0 which is not NaN

// dates
isNaN(new Date());                // false
isNaN(new Date().toString());     // true

// This is a false positive and the reason why isNaN is not entirely reliable
isNaN('blabla');   // true: "blabla" is converted to a number.
                   // Parsing this as a number fails and returns NaN
```

Also, the current numeric options TypeBox provides are based on the standard numeric properties defined within the JSON Schema Draft 6 specification (you can read up on these [here](https://json-schema.org/understanding-json-schema/reference/numeric.html)). So even if a safe `isNaN` option were possible, it still couldn't be added to TypeBox as validating for such criteria would put TypeBox at odds with the JSON Schema specification. 

However, there are plans to enable custom schema validators in future releases of TypeBox. Such custom validators would be able to support custom `isNaN` type validation, but users will need to implement these validators external to TypeBox. See issue https://github.com/sinclairzx81/typebox/issues/244 for a little bit of thinking currently going into this functionality.

Thanks again for the PR, but will close off for now.

Kind Regards
S

PR #261, opened 2022-10-18
support pick from const array

noamokman:
Hey,
A very small fix to support passing a const array to the pick function.
I added a test case to illustrate the use case.
Thanks!


sinclairzx81:
@noamokman Hi, thanks for this!

The update looks ok to me. But would it be possible to apply a similar update to the `Type.Omit()` type also? The `Type.Pick()` and `Type.Omit()` utility types are a pair so would be good to have them updated together.

noamokman:
Sure, on it :)

noamokman:
Done 👍🏻

sinclairzx81:
@noamokman Hey, Awesome :)

Everything looks good to me, will merge this one through and push the updates shortly

Thanks again for the PR, this looks great! :) 
S

sinclairzx81:
@noamokman Updates published on `0.24.47`. Nice work! :)

PR #266, opened 2022-10-26
Allow Integer keys in Record types

sds:
Hello! I recognize the contribution guidelines ask to open an issue first, but I think a PR might help better explain what I'm hoping to accomplish with the following suggestion.

Currently, if you attempt to define a record with an integer key you'll get a type error:

```ts
t = Type.Record(Type.Integer(), ApiUser);
```

```
Argument of type 'TInteger' is not assignable to parameter of type 'TString<string> | TNumber'
```

Considering numbers are allowed as `Record` keys, it seems reasonable to also allow integers.

With that said, I might be missing something, so feel free to close this if this doesn't align.

Thanks for your consideration!

### DIFF
--- a/src/typebox.ts
+++ b/src/typebox.ts
@@ -380,7 +380,7 @@ export interface TPromise<T extends TSchema = TSchema> extends TSchema {
 // Record
 // --------------------------------------------------------------------------
 
-export type TRecordKey = TString | TNumber | TUnion<TLiteral<any>[]>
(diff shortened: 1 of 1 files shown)


sinclairzx81:
@sds Hi! 

Hey thank you the PR, and yeah, this does look like an oversight to me. Excellent find! :)

Happy to merge this through. I'll drop a couple of unit tests around it just to verify before publishing, but the update looks good to me :)

Thanks again!
S

sinclairzx81:
Updates published on `0.24.51`

Good work :)

PR #292, opened 2022-12-18
Structural Object Check

sinclairzx81:
This PR implements a compiler optimization and structural check for `isArray()` when testing object schematics. This optimization only applies to object checks when testing partial objects, but has the following implication when testing array values against object schematics. This PR opts for the structural interpretation of a JavaScript value, not the nominal interpretation (as is the case for Ajv)

### Cases

```typescript
import { TypeCompiler } from '@sinclair/typebox/compiler'

// Case A
{
  const T = Type.Object({ length: Type.Number() })
  const C = TypeCompiler.Compile(T)
  
  C.Check([0, 1, 2]) //  true: because array.length is numeric
}

// Case B
{
  const T = Type.Object({ length: Type.Number() }, { additionalProperties: false })
  const C = TypeCompiler.Compile(T)
  
  C.Check([0, 1, 2]) // false because the keys ['0', '1', '2'] are additional properties.
}
```

### DIFF
--- a/example/index.ts
+++ b/example/index.ts
@@ -10,7 +10,7 @@ import { Type, Kind, Static, TSchema } from '@sinclair/typebox'
 const T = Type.Object({
   x: Type.Number(),
   y: Type.Number(),
-  z: Type.Number()
+  z: Type.Number(),
 })
 
 type T = Static<typeof T>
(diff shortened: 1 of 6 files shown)


sinclairzx81:
Closing due to non-spec alignment.

PR #302, opened 2023-01-07
fix(compiler): concatenate escape string of regex

yeongjet:
To solve #303


sinclairzx81:
@yeongjet Hi, thanks for this. This is a good catch! :)

The fix looks all good to me, will merge this one through.

Many Thanks!
S

sinclairzx81:
@yeongjet Updates on `@sinclair/typebox@0.25.17`

Thanks again!

Issue #260, opened 2022-10-14
Question: Comments on static types


jsbrain:
I was wondering if you think there is a way to have property comments reflected on the `Static<>` type, e.g.:

```ts
const testSchema2 = {
  /** This is some property. */
  someProp: 'test',
};
type TestType2 = typeof testSchema2;
export const testObj2: TestType2 = {
  someProp: 'test'
};
```

So when I "normally" infer a type from an object, I will receive the comments in the intellisense like so:

<img width="343" alt="Screen Shot 2022-10-14 at 10 24 37" src="https://user-images.githubusercontent.com/20225909/195799490-6b0f4f00-f643-46e5-9489-289eb48a660f.png">

But when I do it as a typebox schema:

```ts
const testSchema = Type.Object({
  /** This is some property. */
  someProp: Type.String(),
});
type TestType = Static<typeof testSchema>;
export const testObj: TestType = {
  someProp: 'test',
};
```

And in the inferred type, the comments aren't present anymore:

<img width="579" alt="Screen Shot 2022-10-14 at 10 24 49" src="https://user-images.githubusercontent.com/20225909/195799627-824ac036-80d4-4cc3-a2ef-eb111707a5d0.png">

Any idea if this would be possible?

sinclairzx81:
@jsbrain Hi. I have looked into this before https://github.com/sinclairzx81/typebox/issues/150#issuecomment-1030143472, but didn't find a good approach to make it work. TypeScript is actually able to create mapped objects types where each property retains the properties doc comment, but this only really works for really simple cases.

The issue getting this to work in TypeBox most stems from needing to decompose a object into 4 separate object types, `readonly`, `readonly-optional`, `optional` and `required`. These 4 types are then intersected together to produce the correct output type. This decomposition strategy enables TypeBox to inference properties modifiers correctly `Type.Optional()`, `Type.Readonly()`, etc, but because of the decomposition, the comments applied to the each property are lost.

So, I don't think this is possible (at least not without introducing a fair amount of additional complexity / compute overhead to the current strategy). However the code to handle this inference logic can be found [here](https://github.com/sinclairzx81/typebox/blob/master/src/typebox.ts#L291-L301) if you want to experiment. I should also note that this inference strategy dates back to the very early versions of TypeBox, and there may be alternative approaches possible today...

jsbrain:
Thank you @sinclairzx81 for this detailed explanation, I already expected something like that ...
[...]

Issue #150, opened 2022-02-04
Object properties lose their comments


cjol:
It's often useful to document type properties in TypeScript, so that they show up in intellisense:
![image](https://user-images.githubusercontent.com/1517675/152550341-9bfd74d1-efe5-4db5-af3b-5a9c4dff0a21.png)

Typebox sadly does not preserve field comments:
![image](https://user-images.githubusercontent.com/1517675/152550039-2163caed-7684-4ef5-bcc0-0ef77131518e.png)

I did a bit of digging, because other similar libraries are able to preserve the comments. I think the issue is in the definition of `StaticObject`. I don't understand the project well enough to understand what the `infer U` conditional is doing, but it seems to interfere with TypeScript's ability to forward the comments in the mapped type:
https://github.com/sinclairzx81/typebox/blob/00117295faa4b244eb1fb914f437fd6ff822fe9d/src/typebox.ts#L222

As a toy example, I tried removing it to demonstrate the problem to myself and it works on my small scale. What do you think?

```ts
// adjusted from source code to remove conditional type
type StaticObject<T extends TProperties> =
   { [K in keyof T]: T[K] };

// copied from source code
interface TObject    <T extends TProperties>                   extends TSchema         { $static: StaticObject<T>} 
declare function Object<T extends TProperties>(properties: T): TObject<T> 


type MyModel = Static<typeof MyModel>;
const MyModel = Object({
  /** This is some documentation */
  amount: Type.Number(),
});
declare const item : MyModel



console.log(item.amount);

```
 
![image](https://user-images.githubusercontent.com/1517675/152553145-52551eaf-ebcf-45a5-9a07-4a26d2a8cb1d.png)

sinclairzx81:
@cjol Hi, Would be very much interested in having doc comments appear on object properties, but not sure if it's possible given the way TypeBox extracts property modifiers.

The conditional mapping that happens on `StaticObject` is fairly complicated, but is written this way to ensure property modifiers `default`, `readonly`, `optional` and `readonly optional` are appropriately mapped. The high level view of the mapping works as follows.

```typescript
// ----------------------------------------------------------------------------------------
// We start with TypeBox type T that needs mapping to TypeScript type S (with T
// shown below as a TypeScript type for clarity)
// ----------------------------------------------------------------------------------------

type T = {
  a: string           // default
  b?: string          // optional
  readonly c: string  // readonly
  readonly d?: string // readonly optional
}
[...]

PR #324, opened 2023-02-16
Fix PropertiesReduce typing

stevezhu:
Dug into these issues a bit since I was running into it.
https://github.com/sinclairzx81/typebox/issues/260
https://github.com/sinclairzx81/typebox/issues/150#issuecomment-1030143472

The main issue is the `PropertiesReduce` type. With the change below, IntelliSense comments for objects can now be linked back to their original schema declaration.

These are the fixed types here:
```typescript
export type PropertiesReduce<T extends TProperties, P extends unknown[]> = 
  Readonly<Partial<Pick<{ [K in keyof T]: Static<T[K], P> }, ReadonlyOptionalPropertyKeys<T>>>> &
  Readonly<Pick<{ [K in keyof T]: Static<T[K], P> }, ReadonlyPropertyKeys<T>>> &
  Partial<Pick<{ [K in keyof T]: Static<T[K], P> }, OptionalPropertyKeys<T>>> &
  Required<Pick<{ [K in keyof T]: Static<T[K], P> }, RequiredPropertyKeys<T>>> extends infer R ? { [K in keyof R]: R[K] } : never
```

Please verify that this works and still maintains the existing requirements for types.

Also not sure if you had this type of structure anywhere else in the library, I only verified this for `Type.Object()`.

Feel free to clean up the types if you want it implemented differently and thanks for the great library!

# Reasoning

I haven't verified this in depth, but my intuition on why this fixes the issue is because `Pick` is implemented in the following way:
```typescript
type Pick<T, K extends keyof T> = {
    [P in K]: T[P];
}
```
Because `Pick` has `K extends keyof T`, TypeScript is able to maintain the link between the original keys because it knows that `K` is a key of `T`.

To use required properties as an example:
```typescript
{ [K in RequiredPropertyKeys<T>]: Static<T[K], P> }
```
`RequiredPropertyKeys<T>` can't be verified to `extend keyof T` here because you could easily implement `RequiredPropertyKeys<T>` as any arbitrary type such as `type RequiredPropertyKeys<T> = 'randomKey'`.

You have to instead use something similar to `Pick` that guarantees that the keys in the mapped type are actually keys of `T`.
```typescript
Required<Pick<{ [K in keyof T]: Static<T[K], P> }, RequiredPropertyKeys<T>>>
```

### DIFF
--- a/src/typebox.ts
+++ b/src/typebox.ts
@@ -319,7 +319,7 @@ export type PropertiesReduce<T extends TProperties, P extends unknown[]> =
   Readonly<Partial<Pick<{ [K in keyof T]: Static<T[K], P> }, ReadonlyOptionalPropertyKeys<T>>>> &
   Readonly<Pick<{ [K in keyof T]: Static<T[K], P> }, ReadonlyPropertyKeys<T>>> &
   Partial<Pick<{ [K in keyof T]: Static<T[K], P> }, OptionalPropertyKeys<T>>> &
-  Required<Pick<{ [K in keyof T]: Static<T[K], P> }, RequiredPropertyKeys<T>>> extends infer R ? { [K in keyof R]: R[K] } : never;
+  Required<Pick<{ [K in keyof T]: Static<T[K], P> }, RequiredPropertyKeys<T>>> extends infer R ? { [K in keyof R]: R[K] } : never
 
 export type TRecordProperties<K extends TUnion<TLiteral[]>, T extends TSchema> = Static<K> extends string ? { [X in Static<K>]: T } : never
 
--- a/src/typebox.ts
+++ b/src/typebox.ts
@@ -316,10 +316,10 @@ export type RequiredPropertyKeys<T extends TProperties> = keyof Omit<T, Readonly
 
 // prettier-ignore
 export type PropertiesReduce<T extends TProperties, P extends unknown[]> = 
-  { readonly [K in ReadonlyOptionalPropertyKeys<T>]?: Static<T[K], P> } & 
-  { readonly [K in ReadonlyPropertyKeys<T>]:          Static<T[K], P> } & 
-  {          [K in OptionalPropertyKeys<T>]?:         Static<T[K], P> } & 
-  {          [K in RequiredPropertyKeys<T>]:          Static<T[K], P> } extends infer R ? { [K in keyof R]: R[K] } : never
+  Readonly<Partial<Pick<{ [K in keyof T]: Static<T[K], P> }, ReadonlyOptionalPropertyKeys<T>>>> &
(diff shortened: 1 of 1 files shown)


sinclairzx81:
@stevezhu Hi!

Hey, this PR looks great! Give me a little bit of time to review these updates. I'll need to do a couple of checks over prior revisions of TypeScript just to ensure there's no issues there, and run a couple of inference test cases to ensure recursion doesn't pop out, but overall this PR looks really promising (and the object inference logic does need modernizing as it dates all the way back to the earliest version of the library). 

Will keep you posted! Thanks again!

stevezhu:
Thanks!

Additionally, not sure if the repeated mapping types have any performance hit for TypeScript, but here's an alternative method where the mapped type is only written once.
```typescript
export type PropertiesReduceHelper<T extends TProperties, M extends { [K in keyof T]: unknown }> = 
  Readonly<Partial<Pick<M, ReadonlyOptionalPropertyKeys<T>>>> &
  Readonly<Pick<M, ReadonlyPropertyKeys<T>>> &
  Partial<Pick<M, OptionalPropertyKeys<T>>> &
  Required<Pick<M, RequiredPropertyKeys<T>>>

export type PropertiesReduce<T extends TProperties, P extends unknown[]> =
  PropertiesReduceHelper<T, { [K in keyof T]: Static<T[K], P> }> extends infer R ? { [K in keyof R]: R[K] } : never
```

sinclairzx81:
@stevezhu Hi, have been through the updates (including helper optimization) and I think this PR pure gold :) Excellent work and thank you so much for the taking the time to investigate this aspect of the library! It's really appreciated.

I'm quite keen to go with the optimization you suggested. I'm actually currently on the lookout for good tooling that might help to measure inference performance just to get a comparative measurement between implementations (there's growing awareness to inference DX in the TS community these days, so always keen to improve TB's performance in this respect). There's a few guidelines here https://github.com/microsoft/TypeScript/wiki/Performance which document some aspects, but I do think explicitly inferencing the static as few times as required is a good call.

Have tested as low as `4.1.5` (minimal threshold for TB) and things are basically good to go.

Quick CheckList
- [ ] Apply suggested optimization
- [ ] Run a `npm run format`

Again excellent work! Looking forward to seeing this up on the next release!
Many Thanks!
S

sinclairzx81 (approved):

sinclairzx81:
@stevezhu Hey, might just merge this one through as your PR is exceptionally good :)

I'll apply those suggested optimizations in the mainline before publishing out to NPM. Updates will go out on the next publish (should be within the next few hours or so).

sinclairzx81:
@stevezhu Updates on `0.25.23`.

Thanks again for this amazing contribution! :)

stevezhu:
@sinclairzx81 Nice! Definitely appreciate the focus on performance; that's one of the main reasons why I chose this library.

And thanks for having such a fast response. Happy to see this released!

PR #378, opened 2023-04-12
Fix: '@' character is not alpha

m-ronchi:
currently this fails:
```
TypeCompiler.Compile(Type.Object({ "@foo": Type.String() }))
```
because the `@` character (ascii code: 64) is mistakenly included in the `Character.IsAlpha` check

### DIFF
--- a/src/compiler/compiler.ts
+++ b/src/compiler/compiler.ts
@@ -64,7 +64,7 @@ namespace Character {
     return code === 95
   }
   export function IsAlpha(code: number) {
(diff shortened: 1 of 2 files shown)


sinclairzx81:
@m-ronchi Hi, this is a good catch :) Will merge this one through and publish shortly.

Many Thanks!
S

sinclairzx81:
@m-ronchi Fix published on `0.27.4`.

Thanks again!
S

PR #390, opened 2023-04-17
Record Distributed Keys

sinclairzx81:
This PR implements changes to support distributed keys on Record types. These updates are implemented to remove the implicit `addtiionalProperties: false` constraint applied to Record types, enabling them to be composed with `Type.Intersect`. 

Updates include changes to the `KeyResolver` which now provides a `ResolvePattern` string used for RegEx property tests. This pattern is required in Intersect tests where intersected Records may be combined with Objects with known keys.


sinclairzx81:
Deferring this PR for upstream indexed accessor type implementation.

PR #426, opened 2023-05-05
DefaultAdditionalProperties policy

russell-dot-js:
Changing the default behavior of setting "additionalProperties" to false makes it quite verbose if you want to strip additional properties everywhere (or add a new property that you want include on every schema)

Let me know what kind of coverage you need to get this merged

### DIFF
--- a/src/system/system.ts
+++ b/src/system/system.ts
@@ -52,6 +52,8 @@ export namespace TypeSystem {
   export let AllowNaN: boolean = false
   /** Sets whether `null` should validate for void types. The default is `false` */
   export let AllowVoidNull: boolean = false
+  /** Default value for `additionalProperties` on Object types */
+  export let DefaultAdditionalProperties: Types.TAdditionalProperties = undefined
   // ------------------------------------------------------------------------
   // String Formats and Types
   // ------------------------------------------------------------------------
--- a/src/typebox.ts
+++ b/src/typebox.ts
@@ -29,6 +29,8 @@ THE SOFTWARE.
 // --------------------------------------------------------------------------
 // Symbols
 // --------------------------------------------------------------------------
+import { TypeSystem } from '@sinclair/typebox/system';
+
 export const Modifier = Symbol.for('TypeBox.Modifier')
 export const Hint = Symbol.for('TypeBox.Hint')
 export const Kind = Symbol.for('TypeBox.Kind')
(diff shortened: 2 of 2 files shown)


sinclairzx81:
@russell-dot-js Hi, Thank you for the PR, but I can't merge this through.

The reason is because automatically applying `additionalProperties: false` will break schema composition, and there isn't a clear way to opt out. For example, you cannot intersect two inner object types which have been constrained with additionalProperties false.

```typescript
const T = Type.Intersect([
  Type.Object({ x: Type.Number() }, { additionalProperties: false }),
  Type.Object({ y: Type.Number() }, { additionalProperties: false })
])

Value.Check(T, { x: 1, y: 1 }) // false
```
Automatically applying this constraint (even through configuration) introduces many difficult to resolve problems (and has been attempted previously). To avoid complications and unexpected behavior, TypeBox follows JSON Schema and requires the user to explicitly state `additionalProperties: false` the same as you would on the raw schema (even at the cost of verbosity). Additional information on closed schemas and complications related to composition can be found at the link below.

https://json-schema.org/understanding-json-schema/reference/object.html#extending-closed-schemas

Again, thanks for the PR, but cannot merge through. Will close.
All the best
S

russell-dot-js:
Given your example above, in this case... given two Type.Objects, couldn't Type.Intersect be smart enough to return

```
{
  type: 'object',
  required: ['x', 'y'],
  properties: {
    x: { type: 'number' },
    y: { type: 'number' },
   },
  additionalProperties: false,
}
```

In typescript,
```
type A = {
  x: number;
  additionalProperties: false;
}
type B = {
  y: number;
  additionalProperties: false;
}

type C = A & B;
// produces:
type C = {
  x: number;
  y: number;
  additionalProperties: false;
}
```

This would work even if only A or B had additionalProperties set to false. The only time it wouldn't work is if one is explicitly set to true, while the other is explicitly set to false, which makes sense... you'd have to Omit the property from one or the other prior to intersecting them.

The change above would still produce a valid JSON Schema. I understand the intent of the library is to provide both valid json schemas, as well as valid types, but is there any hard and fast rule that the library must produce json schema that's 1:1 with what the user passed in?

sinclairzx81:
> The change above would still produce a valid JSON Schema. I understand the intent of the library is to provide both valid json schemas, as well as valid types, but is there any hard and fast rule that the change I'm suggesting above breaks?

Unfortunately yes. I had a very lengthy discussion with one of the core JSON Schema contributors on this topic (which you can read about [here](https://github.com/json-schema-org/json-schema-spec/issues/1087)). The main problem is removing the additionalProperties constraint from the schema changes the semantics of that schema, and introduces ambiguity in places which are difficult to reconcile. Note, TypeBox previously implemented default `additionalProperties: false` (around the time of this discussion), however this implementation proved extremely difficult to reconcile for certain types (notably for optional properties, partially constrained objects (ones with the constraint and ones without), and overlapping properties of varying types). All of these things need factoring with the inclusion of a default `additionalProperties: false` constraint.

Ultimately, the issue is that TypeBox would need to "imply" things of the JSON Schema specification that just do not exist (as JSON Schema has no facilities for extending closed schemas), and as such, the decision was made to have TypeBox be a merely a thin layer on top of the JSON Schema, but to never imply things that are not immediately expressible through the specification itself (including making library specific decisions as to how to reconcile the removal of constraints during composition)

As a consequence of this decision, users are responsible to constraining and removing constraints through composition. This can't be automatic (as this could lead to surprising behavior and be potential at odds with future enhancements to the specification)

### TypeSystem.DefaultAdditionalProperties

Just on this. I can't add policies that result in varying schema representations. The current representations are fixed to what they are, with the policies only relating to assertion rules implemented by the type checkers. So while `DefaultAdditionalProperties` wouldn't be able to apply  `additionalProperties` to the schemas (as per this PR), the type checkers could potentially observe this rule and assert the schemas as if `additionalProperties` was specified.

However, as users can still specify things like `additionalProperties: Type.String()`, having a global override to imply `additionalProperties: false` may not be the best idea, as it may lead to potential maintenance problems handling cases where users do need to additional properties. So somewhat apprehensive about a global setting for this. 

None of this is ideal of course, but for the purposes of JSON schema alignment (and keeping things as simple and apparent as possible), requiring users to explicitly configure types with `additionalProperties: false` is necessary (at least for now and the foreseeable future)

Hope this brings a bit of insight as to why this PR was closed
All the best
S

PR #434, opened 2023-05-13
Fix unevaluatedProperties and additionalProperties in nested schemas in TypeCompiler 

gorbak25:
Currently the following schema:
```
Type.Object({nested: Type.Object({}, {additionalProperties: Type.String()})})
```
after being compiled using TypeCompiler fails to validate the following object:
```
{ nested: { foo: 'bar' }}
```
This is due to scoping bugs in the TypeCompiler. After reading the code of the compiler I also found the same bug with the unevaluatedProperties feature.


sinclairzx81:
@gorbak25 Hey, thanks for the PR and good spotting!

Yes, this looks like an oversight as of some recent changes made to the compiler (my bad). The implementation you have looks immediately correct, also thank you for the additional unit tests! Will merge this one through and should be able to publish shortly after. 

Great work, Cheers!
S

sinclairzx81:
@gorbak25 Just published these changes on `0.28.10`

Thanks again! :)
S

PR #436, opened 2023-05-16
chore(docs): add attention for compile method

erfanium:


sinclairzx81:
Hi, thanks for the PR.

Would it be possible to change the copy to the following?

### Update

Use the `Compile(...)` function to compile a type. Note that compilation is an expensive operation that should typically be performed once per type during application start up. TypeBox does not cache previously compiled types, so applications are expected to hold references to each compiled type for the lifetime of the application.

erfanium:
@sinclairzx81 done

sinclairzx81:
Thanks, if you can just remove the newline there, this should be good to merge (should be a single paragraph)

image

erfanium:
@sinclairzx81 done! also you have edit access for other possible little things

sinclairzx81:
@erfanium LGTM, Thanks for the PR! Will merge through shortly.

PR #456, opened 2023-06-10
Update compiler to reuse local functions

melbourne2991:
A small change to the compiler implementation to reuse local functions. May improve the memory footprint but have not benchmarked it. 

Given the following:

```typescript
import { Type } from "src/typebox";
import * as Types from "src/typebox";
import { TypeCompiler } from "src/compiler/compiler";

const shared = Type.Object({
  boo: Type.String(),
}, {
  $id: "shared"
})

const x = TypeCompiler.Code(Type.Object({
  a: shared,
  b: shared,
}))

console.log(x)
```

Output before change (Currently the compiler will only reuse local functions once and then inline the rest):
```typescript
function check_shared(value) {
  return (
    (typeof value === 'object' && value !== null && !Array.isArray(value)) &&
    (typeof value.boo === 'string')
 )
}
return function check(value) {
  return (
    (typeof value === 'object' && value !== null && !Array.isArray(value)) &&
    check_shared(value.a) &&
    (typeof value.b === 'object' && value.b !== null && !Array.isArray(value.b)) &&
    (typeof value.b.boo === 'string')
 )
}
```

Output after change:
```typescript
function check_shared(value) {
  return (
    (typeof value === 'object' && value !== null && !Array.isArray(value)) &&
    (typeof value.boo === 'string')
 )
}
return function check(value) {
  return (
    (typeof value === 'object' && value !== null && !Array.isArray(value)) &&
    check_shared(value.a) &&
    check_shared(value.b)
 )
}
```

### DIFF
--- a/src/compiler/compiler.ts
+++ b/src/compiler/compiler.ts
@@ -364,19 +364,25 @@ export namespace TypeCompiler {
     state_remote_custom_types.set(schema_key, schema)
     yield `custom('${schema[Types.Kind]}', '${schema_key}', ${value})`
   }
-  function* Visit<T extends Types.TSchema>(schema: T, references: Types.TSchema[], value: string): IterableIterator<string> {
+  function* Visit<T extends Types.TSchema>(schema: T, references: Types.TSchema[], value: string, root = false): IterableIterator<string> {
     const references_ = IsString(schema.$id) ? [...references, schema] : references
     const schema_ = schema as any
     // Reference: Referenced schemas can originate from either additional schemas
     // or inline in the schema itself. Ideally the recursive path should align to
     // reference path. Consider for refactor.
-    if (IsString(schema.$id) && !state_local_function_names.has(schema.$id)) {
-      state_local_function_names.add(schema.$id)
+    if (IsString(schema.$id)) {
       const name = CreateFunctionName(schema.$id)
-      const body = CreateFunction(name, schema, references, 'value')
-      PushFunction(body)
-      yield `${name}(${value})`
-      return
+
+      if (!state_local_function_names.has(schema.$id)) {
+        state_local_function_names.add(schema.$id)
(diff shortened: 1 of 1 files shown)


sinclairzx81:
@melbourne2991 Hi, thanks for this PR :) This is a good one.

I've done some local testing this side and everything looks good. I'll merge this one through and just make couple of small updates prior to publishing. This was actually a known issue with the compiler (I had been meaning to get around to resolving this at some point), so it's great you've picked up and provided a solution here.

I'll note, some of these aspects of the compiler (specifically hoisting for types with `$id`) may signal a broader need to refactor these parts of the compiler in the long term. I note some of the recursive logic during generation was a bit non-trivial to solve for the hoisting case (which is largely why it had been deferred), but the conditional `root` check looks like a nice low cost solution in the interim.

Will merge this one through and will notify when published.
Excellent work! :)
S

sinclairzx81:
@melbourne2991 Heya, updates on `0.28.15`

Just made a small additional update to ensure the top most schema is also hoisted if it has an `$id`. This should keep things fairly consistent.

```typescript
const T = Type.Object({
  x: Type.Number(),
  y: Type.Number(),
  z: Type.Number(),
}, { $id: 'T' }) // $id causes hoisting

// produces
function check_T(value) {
  return (
    (typeof value === 'object' && value !== null && !Array.isArray(value)) &&
    (typeof value.x === 'number' && Number.isFinite(value.x)) &&
    (typeof value.y === 'number' && Number.isFinite(value.y)) &&
    (typeof value.z === 'number' && Number.isFinite(value.z))
 )
}
return function check(value) {
  return check_T(value)
}
```
May look at refactoring some of this in subsequent revisions and simplify some of this logic, but for now, this should be good to go :)
Thanks again!
S

PR #459, opened 2023-06-11
fix: update options type in SchemaOptions interface

JS-AK:
After the changes, the Type.Enum now statically types the default value in the options.

image
image

### DIFF
--- a/src/typebox.ts
+++ b/src/typebox.ts
@@ -80,7 +80,7 @@ export type Key = string | number
 // --------------------------------------------------------------------------
 // TSchema
 // --------------------------------------------------------------------------
-export interface SchemaOptions {
+export interface SchemaOptions<T extends { default: any } = any> {
   $schema?: string
   /** Id for this schema */
   $id?: string
@@ -89,7 +89,7 @@ export interface SchemaOptions {
   /** Description of this schema */
   description?: string
   /** Default value for this schema */
-  default?: any
+  default?: T['default']
   /** Example values matching this schema */
   examples?: any
   [prop: string]: any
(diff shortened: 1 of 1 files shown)


sinclairzx81:
@JS-AK Hi!

Thanks for the PR and apologies for the delay in reply on this one (have been somewhat out of action for the past few days). 

At this time, the `default` property is typed as `any?` as the data used to construct the default value may not be necessarily be the same type of that value. For example, consider the following where the `time` property may be initialized lazily and returns the current timestamp `now()`.

```typescript
const T = Type.Object({
  time: Type.Number({ default: () => Date.now() })        // Function != number
})
```
Or where the `content` originates from a http request.
```typescript
const T = Type.Object({
  content: Type.String({ default: async () => await fetch('http:...').then(res => res.text()) })
})
```
TypeBox's default `any?` implementation is somewhat informed by the following description

https://json-schema.org/understanding-json-schema/reference/generic.html

> The default keyword specifies a default value. This value is not used to fill in missing values during the validation process. Non-validation tools such as documentation generators or form generators may use this value to give hints to users about how to use a value. However, default is typically used to express that if a value is missing, then the value is semantically the same as if the value was present with the default value. The value of default should validate against the schema in which it resides, but that isn’t required. (<<< this last past)

Currently, TypeBox will initialize the `default` value using the immediate value (via `Value.Create()` which ideally should be specified as a value matching the expected type), but can't exclude the non-type correct values (i.e. lazy values), or even metadata that leads to a value via other processes (like database id lookup or http request). For this reason, somewhat keen to keep the current definition of `default` as non-typed which allows implementers to build all sorts of value construction mechanisms (not just the ones implemented by TypeBox)

I do actually agree the `default` value should be typed accordingly, but at this stage, still keen to keep things flexible and open as not to implicate scenarios where the `default` property holds metadata or functions to produce values, and where it's not possible to know in advance what that data should be. One possible approach might be to provide inference in the following way, where user defined default values infer as such, but still remain decoupled from the actual type (allowing for a level of indirection between default metadata and the eventual value)

```typescript
const T = Type.String({ default: 123 }) 

type T = (typeof T)['default'] // current:  type T = any  
                               // proposed: type T = 123
```

Hope that brings some insights into why `default` is typed as `any`. Happy to discuss a bit more, but might be a better discussion for an issue thread or discussion thread. 

Again, apologies for the delay on this one. I do like the premise and thinking behind this PR but unfortunately cannot merge for the above reason.
Thanks again
S

PR #463, opened 2023-06-13
Fixed some typos in readme

alexbadm:


sinclairzx81:
@alexbadm Hi :)

Thank you very much for this! Everything looks great here. Will merge this through now and publish on the next revision (this will be 0.28.16 when the next code updates roll through)

Thanks again!
S

PR #473, opened 2023-06-23
Fix/add support numeric records to enum

JS-AK:
Fixed empty generated anyOf: [] of Type.Enum() if passing numeric like values (and ajv throw error when validate jsonSchema with empty anyOf array)

before 

```
Type.Enum({ 1: 1, 2: 2 } as const)          // { anyOf: [], [Symbol(TypeBox.Kind)]: 'Union' }

Type.Enum({ 1: '1', 2: '2' } as const)       // { anyOf: [], [Symbol(TypeBox.Kind)]: 'Union' }
```

after pr changes

```
Type.Enum({ 1: 1, 2: 2 } as const)          // { anyOf: [{ type: 'number', const: 1, [Symbol(TypeBox.Kind)]: 'Literal' }, { type: 'number', const: 2, [Symbol(TypeBox.Kind)]: 'Literal' }], [Symbol(TypeBox.Kind)]: 'Union'}

Type.Enum({ 1: '1', 2: '2' } as const)       // { anyOf: [{ type: 'string', const: '1', [Symbol(TypeBox.Kind)]: 'Literal' }, { type: 'string', const: '2', [Symbol(TypeBox.Kind)]: 'Literal' }], [Symbol(TypeBox.Kind)]: 'Union' }
```

### DIFF
--- a/src/typebox.ts
+++ b/src/typebox.ts
@@ -2403,7 +2403,7 @@ export class StandardTypeBuilder extends TypeBuilder {
   /** `[Standard]` Creates a Enum type */
   public Enum<T extends Record<string, string | number>>(item: T, options: SchemaOptions = {}): TEnum<T> {
     // prettier-ignore
-    const values = globalThis.Object.keys(item).filter((key) => isNaN(key as any)).map((key) => item[key]) as T[keyof T][]
+    const values = globalThis.Object.keys(item).map((key) => item[key]) as T[keyof T][]
     const anyOf = values.map((value) => (typeof value === 'string' ? { [Kind]: 'Literal', type: 'string' as const, const: value } : { [Kind]: 'Literal', type: 'number' as const, const: value }))
     return this.Create({ ...options, [Kind]: 'Union', anyOf })
   }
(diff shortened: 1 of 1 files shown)


sinclairzx81:
@JS-AK Hi, thanks for the PR!

I don't think I can merge this one as numeric enum keys are generally not supported (and there's quite a few downstream tests breaking as they make assumptions on enum keys being `string`). Note, the `Type.Enum()` type is principally there to express TypeScript's `enum` construct (which encodes both keys and values). So supporting both numeric and non-numeric variants may not be possible. Consider.

```typescript
enum Foo { A, B } 

// This is what TypeScript emits for enum where the `isNaN` check needs to filter for numerics and 
// only produce `anyOf: []` for the string key values. Unfortunately, JavaScript makes differentiating
// numeric keys and string keys (via getOwnPropertyKeys() or Keys()) difficult.

console.log(Foo) // { '0': 'A', '1': 'B', A: 0, B: 1 }
```
Because of the way TS encodes enums, I don't think it's possible support both object literal and the enum construct under the same type (with the object literal variant only possible as it partial conforms to the encoded TS construct for enum). I think in cases like this, rather than using `Type.Enum({ ... } as const)`, a more robust implementation would be to use `Type.Union([...Type.Literal(...)])` as this removes any uncertainty with respect to user defined property names.

Will close this PR for now, but happy to discuss things further if you have any follow up questions. I think while there may be a way to implement a heuristic to detect both object literal and enum encodings (it would be complicated), the complexity of implementing this tagged with a lot of uncertainty around user defined properties, I'm thinking it may be best to defer for the time being.

All the best!
S

Issue #500, opened 2023-07-17
Custom errors using TypeCompiler


JimSalies:
First of all, thank you for this amazing lib!

I have been searching through documentation / PR and discussions and didn't find anything about using custom errors with typecompiler.  Is it possible to use custom error messages? If so, how to do it?

sinclairzx81:
@JimSalies Hi,

TypeBox doesn't handle automatic custom error messages internally, rather it requires implementers to remap the error objects yielded on `.Errors()` externally as they need them. This is usually handled by writing a mapping function over the top of `.Errors()` which implements any custom error mapping logic you need. The following code does a quick inline remap to demonstrate how you can extract and remap custom error messages applied to schemas.

```typescript
const T = Type.Object({
  x: Type.Number({ errorMessage: 'X error' }),
  y: Type.Number({ errorMessage: 'Y error' }),
  z: Type.Number({ errorMessage: 'Z error' }),
})

const C = TypeCompiler.Compile(T)

// The failing schema is passed as a property on each `error` object. You can inspect 
// the schema and remap for any custom metadata found on it if necessary. Note, you 
// would typically wrap the following up in a function and use it in your applications
// validation pipeline.
const E = [...C.Errors({ x: '1', y: '2', z: '3' })].map(error => {
  return (typeof error.schema.errorMessage === 'string')
    ? {...error, message: error.schema.errorMessage}
    : {...error }
})

console.log(E) // const E = [  
               //   {
               //     type: 29,
               //     schema: {
               //       errorMessage: 'X error',
               //       type: 'number',
               //       [Symbol(TypeBox.Kind)]: 'Number'
               //     },
               //     path: '/x',
               //     value: '1',
               //     message: 'X error'                      // message mapped here
               //   },
               //   {
               //     type: 29,
               //     schema: {
               //       errorMessage: 'Y error',
               //       type: 'number',
               //       [Symbol(TypeBox.Kind)]: 'Number'
               //     },
               //     path: '/y',
               //     value: '2',
               //     message: 'Y error'                      // message mapped here
               //   },
               //   {
               //     type: 29,
               //     schema: {
               //       errorMessage: 'Z error',
               //       type: 'number',
               //       [Symbol(TypeBox.Kind)]: 'Number'
               //     },
               //     path: '/z',
               //     value: '3',
               //     message: 'Z error'                     // message mapped here
               //   }
               // ]
```
[...]

PR #501, opened 2023-07-18
fix: Object maxProperties error message references wrong value

jayalfredprufrock:


jayalfredprufrock:
I didn't add this to my commit, but you might want to consider using language in the error message to indicate the "inclusive" nature of the comparison, e.g. "Expected object to have _no more than_ X properties".

sinclairzx81:
@jayalfredprufrock Hey thanks :) This is a good catch!

> I didn't add this to my commit, but you might want to consider using language in the error message to indicate the "inclusive" nature of the comparison, e.g. "Expected object to have no more than X properties".

Yeah, I may look at some of the error reporting messages in later versions. To be honest, I'm almost tempted to remove the messages entirely and just work off the `ValueErrorType` enum. I think to support i18n, it may be easier to just have mapping functions that map the error type + schema to multiple languages. Occasionally I get issues posted like https://github.com/sinclairzx81/typebox/issues/500 where users ask about custom error reporting, so just making the mapping a requirement would solve both custom and i18n requirements (but something to look into for another day) :)

Thanks again for picking up on this, will merge across and publish a revision shortly.

jayalfredprufrock:
FWIW, I support getting rid of the error messages altogether. They have come in handy in the past, but they could be handled just as easily in userland and as you say, that would open the door for i18n as well.

PR #542, opened 2023-08-24
AOT compile regexp for records

smitssjors:
```typescript
TypeCompiler.Compile(Type.Record(Type.String(), Type.Any())).Code()
``` 
Gives
```typescript
const local_0 = new RegExp(/^(.*)$/)
// ...
``` 
But I belive that literal notoation would be better. Thus
```typescript
const local_0 = /^(.*)$/
```


sinclairzx81:
@smitssjors Hi, thanks for the PR

This looks good to me. I see I'm actually using this string embedding for the other types (String, TemplateLiteral), so it makes sense to use it for Record types also :)

Will merge this one through and will publish shortly.
Many thanks!
S

sinclairzx81:
@smitssjors Hiya,

Updates published on `@sinclair/typebox@0.31.3`

Thanks again for the PR!
Cheers
S

Issue #564, opened 2023-08-30
A record with an enum as key resolves to TObject<{}>


dippi:
# Issue

```typescript
enum E {
  A = 'A',
  B = 'B',
  C = 'C',
};
const T = Type.Record(Type.Enum(E), Type.Number());

```
`T` has type `TObject<{}>` instead of `TObject<{ A: TNumber; B: TNumber; C: TNumber }>`

This is because `Type.Enum(E)` is equivalent to `TUnion<E[]>` instead of `TUnion<['A', 'B', 'C']>` and in `TRecordFromUnionRest` it doesn't match `K extends [infer L, ...infer R]`.

https://github.com/sinclairzx81/typebox/blob/010ff1840c7ab43d3df6a52f922e365184635217/src/typebox.ts#L653-L657

# Possible fix

I tried adding `K extends TEnum<infer S> ? TObject<{ [K in keyof S]: T }> :` as first case of `TRecordResolve` and it seems to be working as expected.

https://github.com/sinclairzx81/typebox/blob/010ff1840c7ab43d3df6a52f922e365184635217/src/typebox.ts#L671-L679

# Regression tests

I'd be happy to make a PR, but I had an issue writing a regression test.

I tried to add the example above in [/test/static/record.ts](https://github.com/sinclairzx81/typebox/blob/18d1cf7ac57c8b5f0292cb02ddbd1707e86c8c4c/test/static/record.ts) with the assertion `Expect(T).ToStatic<Record<E, string>>();`, but it succeeds even though it shouldn't.

The reason is that `.ToStatic<U>()` only checks that `U` extends `Static<T>` instead of checking equality (and `{}` is super-type of `Record` so it always succeeds).

In another project I took inspiration from [gvergnaud/ts-pattern](https://github.com/gvergnaud/ts-pattern/blob/435284ac08cd0c4c1361fd59ae936bd1ece7c85b/src/types/helpers.ts#L67-L71) to assert type equality and I could use it for this specific case.

However I think the best case scenario would be to somehow evolve `.ToStatic` so that the whole test suite would benefit from it, but I don't know how.

Let me know if you have any suggestion.

# Additional context

We updated from `0.29.4` to `0.31.5`, but I also checked later versions and the issue persists.

sinclairzx81:
@dippi Hi, Thanks for reporting! (and for the detailed write up)

[...]

PR #567, opened 2023-08-31
Map Enum as Union Type

sinclairzx81:
This PR Implements a new inference strategy for Enum types. It removes the `TEnum` interface and opts for direct `TUnion` mapping via alias type. This update enables `TEnum` to be used as Record Key, as well as with other type compositions through direct union mapping.

Related Issue https://github.com/sinclairzx81/typebox/issues/564


sinclairzx81:
@dippi Heya, Unfortunately, had to revert these changes as the enum union expansion turned out to be breaking for users with large enums (items > 32). I don't think there's an immediate fix to this (as the UnionToTuple turned out to be the primary culprit  as it really doesn't scale very well).

I would be keen to see if there is a way to retain the union mapping for enum (as the mapping was actually ideal and simplified a lot of internal inference), but do achieve, it would require a very optimal UnionToTuple implementation, or some other inference strategy that would yield the `[TLiteral<...>, TLiteral<...>, TLiteral<...>]` structure without the type instantiation issues.

Last good revision of the implementation can be found https://github.com/sinclairzx81/typebox/releases/tag/0.31.13. If you would like to help experiment, any insights to try and optimize this piece would be hugely welcome :) If there's a way to get this functionality back in, would be looking at reintroducing on a minor revision (so 0.32.x)

All the best
S

dippi:
Hi @sinclairzx81, sorry to hear that the new implementation didn't work out.

I don't have a good idea on how it could be optimized, but I'll think about it in the next few days ;)

PR #569, opened 2023-09-02
More strict type assertions

dippi:
# Alternatives

In the PR there are three possible implementations of constraints:
1. Strict equality
2. Mutually assignable
3. A complex set of conditions

# The good news

All of them fix the original issue to write a regression test for a record with an enum as key.
The following assertions fails on master and succeeds on the enum branch:

https://github.com/sinclairzx81/typebox/blob/23baf023d0d708bf6fc2d706a9796d91e6eb306c/test/static/record.ts#L81

# The bad news

A common shortcoming to all the alternatives (including the current implementation) is that `.ToStatic<never>()` always succeeds, because `never` is assignable to any type, and I don't think there is a way around it unless we change completely the `Expect(x).ToStatic<Y>()` pattern.

E.g. this assertion is completely useless:

https://github.com/sinclairzx81/typebox/blob/18d1cf7ac57c8b5f0292cb02ddbd1707e86c8c4c/test/static/exclude.ts#L5-L6

If this is a deal breaker you can stop reading here and we should brainstorm some way to rewrite static tests.

# The issue with strict equality

I like strict equality because catches even subtle changes, however the following case gives error even though it shouldn't and I can't make it work:

https://github.com/sinclairzx81/typebox/blob/18d1cf7ac57c8b5f0292cb02ddbd1707e86c8c4c/test/static/omit.ts#L89-L94

`Static<typeof P>` is exactly as specified in `.ToStatic<...>`, but somehow they are considered different, because the type literal is simplified to `{ x: number, y: number }` while `Static<typeof P>` somehow preserves `({} | {} | {})` as is.

See also:
- https://github.com/microsoft/TypeScript/issues/51019
- https://dev.to/tylim88/typescript-caveat-6-non-collapsing-object-unions-2i87

The only workaround I can't think of is to treat this test as a special case and just check for assignability.

# The issue with mutual assignability

Strict equality showed that `T.properties` is `any` and these assertions passed even though it is likely not what they were meant to be:

https://github.com/sinclairzx81/typebox/blob/18d1cf7ac57c8b5f0292cb02ddbd1707e86c8c4c/test/static/intersect.ts#L35-L36

Unfortunately `any` is mutually assignable to every type except for `never`, so this issue would go undetected.

# The issue with the complex set of conditions

With this last attempt I tried to cover some corner cases of mutual assignability. It detects when the inferred type is `any` unless explicitly expected to be so.

However the following case is caught only by strict equality:

https://github.com/sinclairzx81/typebox/blob/23baf023d0d708bf6fc2d706a9796d91e6eb306c/test/static/record.ts#L20-L23

I can't tell if there could be other common cases that would go undetected.

# Final considerations

I'd go for strict equality, if we find acceptable to have a special case for the problematic test. Otherwise I think the third option strikes the best balance, but the implementation is way more complex and I'm not sure it's justified.

### DIFF
--- a/test/static/assert.ts
+++ b/test/static/assert.ts
@@ -1,6 +1,7 @@
 import { Static, StaticDecode, StaticEncode, TSchema } from '@sinclair/typebox'
 
 declare const unsatisfiable: unique symbol
+// Warning: `never` and `any` satisfy the constraint `extends Expected<...>`
 type Expected<_> = { [unsatisfiable]: never; }
 
 type ConstrainEqual<T, U> = (
@@ -13,9 +14,35 @@ type ConstrainEqual<T, U> = (
 type CircularHelper<T, U> = [T] extends U ? T : Expected<T>
 type ConstraintMutuallyExtend<T, U> = CircularHelper<T, [U]>
 
-// Comment one or the other to see the difference
+type If<T, Y, N> = T extends true ? Y : N
+type And<T, U> = If<T, U, false>
+type Or<T, U> = If<T, true, U>
+type Not<T> = If<T, false, true>
+
+type IsAny<a> = 0 extends 1 & a ? true : false
+type Extends<T, U> = [T] extends [U] ? true : false
+
+// If U is never, there's nothing we can do
+type ComplexConstraint<T, U> = If<
+  // If U is any, we can't use Expect<T> or it would satisfy the constraint
+  And<Not<IsAny<T>>, IsAny<U>>,
+  never,
+  If<
+    Or<
+      // If they are both any we are happy
+      And<IsAny<T>, IsAny<U>>,
+      // If T extends U, but not because it's any, we are happy
+      And<Extends<T, U>, Not<IsAny<T>>>
+    >,
+    T,
+    Expected<T>
+  >
+>
+
+// Uncomment one of the lines to see the difference
 // type Constraint<T, U> = ConstraintMutuallyExtend<T, U>
-type Constraint<T, U> = ConstrainEqual<T, U>
+// type Constraint<T, U> = ConstrainEqual<T, U>
+type Constraint<T, U> = ComplexConstraint<T, U>
 
 export function Expect<T extends TSchema>(schema: T) {
   return {
--- a/test/static/assert.ts
+++ b/test/static/assert.ts
@@ -20,8 +20,9 @@ type And<T, U> = If<T, U, false>
 type Or<T, U> = If<T, true, U>
 type Not<T> = If<T, false, true>
 
-type IsAny<a> = 0 extends 1 & a ? true : false
 type Extends<T, U> = [T] extends [U] ? true : false
+type IsAny<T> = 0 extends 1 & T ? true : false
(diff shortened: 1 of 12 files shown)


sinclairzx81:
@dippi Hi, this looks good :)

Would it be possible to retarget from `master` to the `enum` branch?

dippi:
Hi @sinclairzx81,

I updated the description with a general explanation of the tradeoffs. Let me know what you think about it.

I also changed the base branch to `enum`, but I chose not to rebase / merge yet, so we can see that the `TEnum` regression test fails before we fix it in the other branch.

sinclairzx81:
@dippi Hi! Thanks for this. This is looking great :)

### Never

I wouldn't worry too much about `.ToStatic<never>()` (as the original tests would also pass against `never`). I'm still happy to roll these updates through as they would be doing a better job than the original assertions, and an incremental refinement is better than no refinement. 

Did you have some thoughts on reversing the generic assertions to capture for never?

### Intersect

> Strict equality showed that T.properties is any and these assertions passed even though it is likely not what they were meant to be:

Yeah, the Intersect case looks like a bug (which it's great the updated asserts picked up on). Can update to the following.

```typescript
const T: TIntersect<[TObject<{ // Assert Type.Intersect Return
    A: TOptional<TString>;
}>, TObject<{
    B: TString;
}>]> = Type.Intersect([A, B])

Expect(T).ToStatic<{
  A?: string,
  B: string
}>()
```
### Record

For the failing Record test, I think it would be good to rebase to `enum` and update the asserts against the new `TEnum` alias type.

```typescript
  enum E {
    A = 'X', // Let's use values different from the enum keys.
    B = 'Y', 
    C = 'Z',
  }
  const T = Type.Record(Type.Enum(E), Type.Number());
  Expect(T).ToStatic<{
    [E.A]: number,
    [E.B]: number,
    [E.C]: number,
  }>();
```
### Constraints

I'm happy with all the constraint implementations there :) (even if the other two are unused today, they may be used for additional / specialized assertions in future, and they are a good reference point). Using the ComplexConstraint as a default should be all good.

Are you able to rebase to `enum`? My preference would be to have the updated asserts passing tests against the `enum` branch before merging, but if you just want to rebase, I can fix up the tests post merge. 

Great work! :)
S

dippi:
Eureka! I think I found a reasonable way to have assertions with `never`.

Also I rebased onto `enum` and fixed all the assertions.

sinclairzx81:
@dippi Hi!

This looks all good to me :) Are you happy for me to merge this one in?

sinclairzx81:
Oh, also, can you just run a typescript test with

```
$ npm run test:typescript
```
This just runs the tests across a few revisions (back to 4.9.5)

dippi:
> Are you happy for me to merge this one in?

Yep, ready!

> Oh, also, can you just run a typescript test with `$ npm run test:typescript`

Ok, I just did and they all pass :tada:

sinclairzx81:
@dippi Awesome!

Will merge through :) Thanks for the PR! This is a great enhancement 

Will aim to have this published on the next revision tomorrow.
All the best
S

PR #593, opened 2023-09-13
Fix/check array with unique items/non latin characters

JS-AK:
Previous version works only for latin characters(exact for first 256 charCodes) and when u check array with objects (!NB with option { uniqueItems: true }) 

with **non latin** strings u got exception on Value.Check() 

TypeError: Cannot mix BigInt and other types, use explicit conversions
      at FNV1A64 (target\test\runtime\index.js:1786:29)
      at StringType (target\test\runtime\index.js:1744:5)
      at Visit (target\test\runtime\index.js:1776:12)
      at ObjectType (target\test\runtime\index.js:1738:5)
      at Visit (target\test\runtime\index.js:1774:12)
      at Hash (target\test\runtime\index.js:1791:3)
      at hashFunction (target\test\runtime\index.js:3291:14)
      at eval (eval at Compile (target\test\runtime\index.js:3275:41), <anonymous>:7:87)
      at TypeCheck.check [as checkFunc] (eval at Compile (target\test\runtime\index.js:3275:41), <anonymous>:7:181)
      at TypeCheck.Check (target\test\runtime\index.js:2773:17)
      at Ok (target\test\runtime\index.js:5089:20)
      at Context.<anonymous> (target\test\runtime\index.js:5289:5)
      at process.processImmediate (node:internal/timers:476:21)

finally fixed them and added test

### DIFF
--- a/src/value/hash.ts
+++ b/src/value/hash.ts
@@ -57,7 +57,7 @@ export enum ByteMarker {
 // --------------------------------------------------------------------------
 let Accumulator = BigInt('14695981039346656037')
 const [Prime, Size] = [BigInt('1099511628211'), BigInt('2') ** BigInt('64')]
-const Bytes = Array.from({ length: 256 }).map((_, i) => BigInt(i))
+const Bytes = Array.from({ length: 2 ** 16 }).map((_, i) => BigInt(i)) // 2 ** 16 is max for charCodeAt() https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charCodeAt
 const F64 = new Float64Array(1)
 const F64In = new DataView(F64.buffer)
 const F64Out = new Uint8Array(F64.buffer)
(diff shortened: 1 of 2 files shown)


JS-AK:
image

sinclairzx81:
@JS-AK Hi, Thanks for the PR.

Unfortunately, the `Bytes` lookup table is currently too big (this table should only contain values between 0-255), but this implementation is generating 65536 elements as given by the `2 ** 16` which is really large and not representative of actual byte values; as (0-65535) is out of range.

```typescript
const Bytes = Array.from({ length: 2 ** 16 }).map((_, i) => BigInt(i))
//                                 ^ too large
```
Hashing operations are limited to byte wise hashing only, however variable length operations run across 1 byte and 2 byte  (short) big int values may cause issues, so would be interested in finding an alternate solution that retains the existing `Bytes` lookup and computing the Unicode byte values derived from the 16-bit value given by `chatCodeAt(x)`.

sinclairzx81:
@JS-AK Heya,

Thanks for reporting this issue (and the PR), but have quickly published an alternate fix on 0.31.15 which derives the bytes from the number returned on `chatCodeAt(x)` and retains the original `Bytes` array. This is generally going to be a bit more memory friendly. 

Thanks again
S

PR #684, opened 2023-12-05
Add string format type : uri

feggaa:


feggaa:
Yes, you are right
I didn't notice it was present because I added it for verification in Elysia and didn't realize its existence in Typebox

 I apologize for any inconvenience.

Issue #336, opened 2023-03-04
Representing JSON values.


maxgurewitz:
Hello, I'm trying to represent the following type,

```typescript
export type JSONValue =
  | string
  | number
  | null
  | boolean
  | { [x: string]: JSONValue }
  | JSONValue[];
```

However, typebox is missing support for index signatures for `{ [x: string]: JSONValue }`. I can't use `Record`'s instead,

```typescript
export const JSONValue = Type.Recursive((Node) =>
  Type.Union([
    Type.Null(),
    Type.String(),
    Type.Number(),
    Type.Boolean(),
    Type.Record(Type.String(), Node),
    Type.Array(Node),
  ])
);

export type JSONValue = Static<typeof JSONValue>;
```

because I get the error `Type instantiation is excessively deep and possibly infinite`.

Any tips? Thank you!

maxgurewitz:
Actually going to move to discussions.

PR #735, opened 2024-01-18
Fix some infinite recursion with static types

ghalle:
This fixes the record type and the array type so that they are less likely to cause infinite recursion.

The `JSONValue` type in #336 is now usable without causing the `Type instantiation is excessively deep and possibly infinite` error.

The record fix was pretty straightforward, simply inlining the logic of `Record` directly in the `RecordStatic` type.
The array fix however is a bit weirder, I'm not quite sure why it works but it does somehow.

No tests were added, I took a brief look at [/test/static/recursive.ts](https://github.com/sinclairzx81/typebox/blob/0c0da45394e5ecb043a6fc6fee7d24a13bc9e4fa/test/static/recursive.ts) but wasn't quite sure how to add a test for this that made sense.

### DIFF
--- a/src/type/array/array.ts
+++ b/src/type/array/array.ts
@@ -27,7 +27,7 @@ THE SOFTWARE.
 ---------------------------------------------------------------------------*/
 
 import { CloneType } from '../clone/type'
-import { Ensure } from '../helpers'
+import { Ensure } from '../helpers/index'
 import type { SchemaOptions, TSchema } from '../schema/index'
 import type { Static } from '../static/index'
 import { Kind } from '../symbols/index'
--- a/src/type/array/array.ts
+++ b/src/type/array/array.ts
@@ -26,10 +26,11 @@ THE SOFTWARE.
 
 ---------------------------------------------------------------------------*/
 
-import type { TSchema, SchemaOptions } from '../schema/index'
+import { CloneType } from '../clone/type'
+import { Ensure } from '../helpers'
+import type { SchemaOptions, TSchema } from '../schema/index'
 import type { Static } from '../static/index'
 import { Kind } from '../symbols/index'
-import { CloneType } from '../clone/type'
 
 export interface ArrayOptions extends SchemaOptions {
   /** The minimum number of items in this array */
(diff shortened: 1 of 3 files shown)


sinclairzx81:
@ghalle Hi, thanks for the PR

Everything looks pretty good here, but will want to get a static test in. I think it should be ok to just use the code from issue https://github.com/sinclairzx81/typebox/issues/336 and just call static on the type, quick snippet below. (for `test/static/recursive.ts`)

```typescript
// issue: https://github.com/sinclairzx81/typebox/issues/336
{
export const T= Type.Recursive((Node) =>
  Type.Union([
    Type.Null(),
    Type.String(),
    Type.Number(),
    Type.Boolean(),
    Type.Record(Type.String(), Node),
    Type.Array(Node),
  ])
)

type T= Static<typeof T>
}
```
Any issues should be picked up on the static tests. Also, just run the following npm commands to make sure everything is all good.

```
$ npm run test               # tests static and runtime
$ npm run test:typescript    # tests static across previous and next versions of typescript
$ npm run build              # builds and asserts the esm import paths are ok
```
> The array fix however is a bit weirder, I'm not quite sure why it works but it does somehow.

Yeah, the array fix is mysterious (looks like a TS quirk to me), but lifting types out into their own thing has been a fix for a couple of things in the past (also the type signature may help TS narrow correctly for the type). 

Good work!

ghalle:
The tests were passing but caught an issue with a missing `/index` at the end of one of the import by running the build command.
Should be good now.

image

sinclairzx81:
@ghalle Hi, apologies for the delay on this (have been somewhat pre-occupied with some other work recently).

Thanks for these amendments. Everything is looking good here, but did you want to quickly fix up small convention issue that was noted above.

```typescript
Evaluate<{ [Key in Assert<Static<K>, PropertyKey>]: Static<T, P>; }>

Evaluate<{ [_ in Assert<Static<K>, PropertyKey>]: Static<T, P>; }>
//          ^ unused Key should be _
```
Once this is updated, should be able to merge through.

sinclairzx81:
@ghalle Awesome thanks! :)

Will merge this through and publish later this afternoon with a couple of other fixes on 0.32.12. Again, apologies for the delay.

> Fixed it, but I can't seem to find where this was mentioned previously?

Hmm, it's possible I didn't publish the comment as a review (it's still showing this side). I don't use many of these features of GH, but had assumed it would be visible, my mistake...

image

Thanks again for the PR, this is a good one :)

All the best!
S

PR #755, opened 2024-02-12
Fix (#754) - Allow decimal values in multipleOf option for Number types

### LINKED ISSUE #754: JS floating point precision bug causes multipleOf validation to fail
Reproduction:
```ts
import { Type } from '@sinclair/typebox';
import { TypeCompiler } from '@sinclair/typebox/compiler';

const multipleOf = 0.1;
const schema = Type.Number({ multipleOf });
const value = 1;

const compiler = TypeCompiler.Compile(schema);

console.log(compiler.Errors(value).First());
// {
//   type: 40,
//   schema: { multipleOf: 0.1, type: 'number', [Symbol(TypeBox.Kind)]: 'Number' },
//   path: '',
//   value: 1,
//   message: 'Expected number to be a multiple of 0.1'
// }
```

This is caused by floating point precision issues in JS when using the modulus (`%`) on decimals, as is being done [here](https://github.com/sinclairzx81/typebox/blob/master/src/errors/errors.ts#L369).

Example:

```ts
// value % multipleOf;
console.log(0 === 1 % 0.1); // 0.099…5
// false
```

A potential fix that avoids additional libraries:

```ts
// Number.isInteger(value / multipleOf);
console.log(Number.isInteger(1 / 0.1));
// true
```

apancutt:
Allow `Number` types to be configured with a decimal `multipleOf` value (fixes #754).

Validation for `BigInt`, `Integer` and `Date` types were unmodified as these should only ever be dealing with modulus-safe integers.

### DIFF
--- a/src/errors/errors.ts
+++ b/src/errors/errors.ts
@@ -366,7 +366,7 @@ function* FromNumber(schema: TNumber, references: TSchema[], path: string, value
   if (IsDefined<number>(schema.minimum) && !(value >= schema.minimum)) {
     yield Create(ValueErrorType.NumberMinimum, schema, path, value)
   }
-  if (IsDefined<number>(schema.multipleOf) && !Number.isInteger(value / schema.multipleOf)) {
+  if (IsDefined<number>(schema.multipleOf) && !IsDecimalSafeMultipleOf(value, schema.multipleOf)) {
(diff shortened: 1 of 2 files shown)


apancutt:
See https://github.com/sinclairzx81/typebox/discussions/757

PR #777, opened 2024-03-08
fix: rethrow TypeBoxError instead of creating a new one

aleclarson:
Valuable information is lost by creating a TransformEncodeError/TransformDecodeError instead of rethrowing the error when it‘s a TypeBoxError, because that error may be a TransformEncodeCheckError/TransformDecodeCheckError, which have an `error` property that tells the caller which property path is the cause.

### DIFF
--- a/src/value/transform/decode.ts
+++ b/src/value/transform/decode.ts
@@ -75,6 +75,7 @@ function Default(schema: TSchema, value: any) {
   try {
     return IsTransform(schema) ? schema[TransformKind].Decode(value) : value
   } catch (error) {
+    if (error instanceof TypeBoxError) throw error
     throw new TransformDecodeError(schema, value, error)
   }
 }
--- a/src/value/transform/encode.ts
+++ b/src/value/transform/encode.ts
@@ -74,6 +74,7 @@ function Default(schema: TSchema, value: any) {
   try {
     return IsTransform(schema) ? schema[TransformKind].Encode(value) : value
   } catch (error) {
+    if (error instanceof TypeBoxError) throw error
     throw new TransformEncodeError(schema, value, error)
   }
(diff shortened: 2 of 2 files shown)


aleclarson:
The existing `throw new TransformDecodeError` statement would still be used for non-TB errors (from a user's custom decode function). We just don't want to wrap TB errors with `TransformDecodeError` since that makes it impossible to get any metadata attached to the original `TransformDecodeCheckError` (in my case, its `error: ValueError` property is of particular interest).

sinclairzx81:
@aleclarson Heya

I think this one can be closed off. I've updated the Transform Error types on 0.32.16 to include the exception that was thrown. I've also updated the Decode/Encode exceptions specifically to include a `path`. Let me know if the updates help out here.


```typescript
// Errors
// thrown externally
// prettier-ignore
export class TransformDecodeCheckError extends TypeBoxError {
  constructor(
    public readonly schema: TSchema, 
    public readonly value: unknown, 
    public readonly error: ValueError // note: path derived from ValueError
  ) {
    super(`Unable to decode value as it does not match the expected schema`)
  }
}
// prettier-ignore
export class TransformDecodeError extends TypeBoxError {
  constructor(
    public readonly schema: TSchema, 
    public readonly path: string, // added: path to where Decode callback failed
    public readonly value: unknown, 
    public readonly error: Error,  // added: the exception that was thrown
  ) {
    super(error instanceof Error ? error.message : 'Unknown error')
  }
}
```
I do think it might be good to try and unify these two errors at some point (at least get the path properties lining up), but this update just augments the existing error types to include additional information. Happy to review general API DX updates with respect to error handling via GH issues/discussion (can discuss things there first)

Thanks again!

PR #781, opened 2024-03-09
fix: avoid Errors call for nested Encode/Decode

### LINKED ISSUE #780: Add JSON path tracking to Encode/Decode
When using `Transform(…).Decode(…).Encode(…)`, it would be great if JSON paths were forwarded through to the `Decode/Encode` callbacks. Then I would be able to thread the JSON path through to any nested `Value.Encode/Decode` calls.

Here's an example of what I'm doing:
```ts
const JsonTransform = <T extends TSchema>(schema: T, options?: ArrayOptions) =>
  Transform(t.String(options))
    .Decode(value => {
      return Value.Decode(schema, JSON.parse(value))
    })
    .Encode(value => {
      return JSON.stringify(Value.Encode(schema, value))
    })

types.GinkgoObject = <T extends TProperties>(
  properties: T,
  options?: SchemaOptions
) => JsonTransform(t.Object(properties, options))

types.GinkgoArray = <T extends TSchema>(schema: T, options?: ArrayOptions) =>
  JsonTransform(t.Array(schema, options))
```

These types are being used for JS<->SQLite marshalling.

## The Problem

When a nested encode/decode fails, the error doesn't have the full JSON path in it, making it hard to debug.

aleclarson:
Use a stack for Encode/Decode to detect when it‘s a nested call. If nested, the `Errors` function is **not** called, since the produced error would only be a fraction of the full JSON path. This change is beneficial for certain `Transform` usage, as described in #780.

Closes #780

### DIFF
--- a/src/value/transform/decode.ts
+++ b/src/value/transform/decode.ts
@@ -63,7 +63,7 @@ export class TransformDecodeCheckError extends TypeBoxError {
   }
 }
 export class TransformDecodeError extends TypeBoxError {
-  constructor(public readonly schema: TSchema, public readonly value: unknown, error: any) {
+  constructor(public readonly schema: TSchema, public readonly value: unknown, error?: any) {
     super(`${error instanceof Error ? error.message : 'Unknown error'}`)
   }
 }
--- a/src/value/transform/encode.ts
+++ b/src/value/transform/encode.ts
@@ -62,7 +62,7 @@ export class TransformEncodeCheckError extends TypeBoxError {
   }
 }
 export class TransformEncodeError extends TypeBoxError {
-  constructor(public readonly schema: TSchema, public readonly value: unknown, error: any) {
+  constructor(public readonly schema: TSchema, public readonly value: unknown, error?: any) {
     super(`${error instanceof Error ? error.message : 'Unknown error'}`)
   }
 }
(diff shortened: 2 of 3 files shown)


sinclairzx81:
@aleclarson Hey, thanks again for these PRs!

Give me a few days to review this and https://github.com/sinclairzx81/typebox/pull/777 as I'm currently unavailable this weekend. Will give this PR some thought, there may be a way to propagate the path into Encode/Decode, but might require an update to the TransformEncode, TransformDecode internal functions (making them work a bit like errors), but will need to review when I'm back.

I do think the Transform functions require a bit of rethinking in terms of error handling, and there may be cross over between this PR and https://github.com/sinclairzx81/typebox/pull/777.  Will try get back to you early next week.

Thanks again!

sinclairzx81:
@aleclarson Hiya,

I think this PR may be resolved by https://github.com/sinclairzx81/typebox/pull/791 where the thrown Error was added to the `Transform{Encode|Decode}Error` type which is now accessible via the `error` property. https://github.com/sinclairzx81/typebox/pull/791 also adds the `path` to this error type indicating which interior value caused the error to be thrown during Encode / Decode.

I think this resolves the original linked issue on https://github.com/sinclairzx81/typebox/issues/780 (which has since been closed, but relates specifically to obtaining error paths). However, can reopen and reinvestigate if I'm missing something with respect to the details around `decodeStack` provided on this implementation. If so, let's discuss via GH issues/discussions!

Thanks again for the PR, will close off this one for now.
All the best!
S

PR #782, opened 2024-03-09
fix: check if Union type is optional during Encode

aleclarson:
Since the “optionality” of a Union type is not inherited by its anyOf types, we need to bail out on `undefined` values when the Union is optional.

### DIFF
--- a/src/value/transform/encode.ts
+++ b/src/value/transform/encode.ts
@@ -171,6 +171,9 @@ function FromTuple(schema: TTuple, references: TSchema[], value: any) {
 }
 // prettier-ignore
 function FromUnion(schema: TUnion, references: TSchema[], value: any) {
+  if (IsOptional(schema) && value === undefined) {
+    return undefined
+  }
   // test value against union variants
   for (const subschema of schema.anyOf) {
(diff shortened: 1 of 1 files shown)


sinclairzx81:
@aleclarson Hi,

I like the idea of bailing out of checks and additional processing of values, but may need a bit more context on this. Note that Optional type is only applicable to PropertKeys of Objects where optionality is determined via `key in value` checks (as opposed to `value === undefined`) so this logic might make more sense added to the `FromObject` method (where key enumeration happens). Also note that ability to bail out on Optional may be applicable to all types (so not limited to Union), which also suggests updates to FromObject.

Can you provide a bit more information (mostly code examples)

1. An example of how the current logic breaks from a end users standpoint (union only)
2. An example / reproduction of where the logic might be observed from a end users standpoint (union only)

Just need to get my bearings on usage patterns before proceeding on this one.

sinclairzx81:
@aleclarson Hey might close off this PR as the functionality seems to be working as expected here (and noting some the finer details around `undefined` and `optional` properties). 

Generally, I think everything is fine here, but if you can draft up a repro and submit as an issue, can maybe take another things.

Thanks again!

Issue #808, opened 2024-03-25
Regression bug 0.32.19


mikicho:
I have encountered a bug when using version 0.32.19, but version 0.32.15 works without any issues.

<blockquote>
➜  bcf-typebox git:(Michael/upgrade-all) ✗ npx tsc
node_modules/@sinclair/typebox/build/import/value/delta/delta.d.mts:4:37 - error TS2307: Cannot find module 'src/type/object/object.mjs' or its corresponding type declarations.

4 export declare const Insert: import("src/type/object/object.mjs").TObject<{
                                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/@sinclair/typebox/build/import/value/delta/delta.d.mts:10:37 - error TS2307: Cannot find module 'src/type/object/object.mjs' or its corresponding type declarations.

10 export declare const Update: import("src/type/object/object.mjs").TObject<{
                                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/@sinclair/typebox/build/import/value/delta/delta.d.mts:16:37 - error TS2307: Cannot find module 'src/type/object/object.mjs' or its corresponding type declarations.

16 export declare const Delete: import("src/type/object/object.mjs").TObject<{
                                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/@sinclair/typebox/build/import/value/delta/delta.d.mts:21:35 - error TS2307: Cannot find module 'src/type/union/union-type.mjs' or its corresponding type declarations.

21 export declare const Edit: import("src/type/union/union-type.mjs").TUnion<[import("src/type/object/object.mjs").TObject<{
                                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Found 4 errors in the same file, starting at: node_modules/@sinclair/typebox/build/import/value/delta/delta.d.mts:4
</blockquote>

tsconfig.json
```json
{
  "compilerOptions": {
    "lib": ["ESNext", "DOM"],
    "moduleResolution": "NodeNext",
    "module": "NodeNext",
    "target": "ESNext",
    "resolveJsonModule": true,
    "incremental": true,
    // Tells TypeScript to read JS files, as
    // normally they are ignored as source files
    "allowJs": true,
    "checkJs": true,
    "strict": true,
    // Generate d.ts files
    "declaration": true,
    // This compiler run should
    // only output d.ts files
    "emitDeclarationOnly": true,
    // Types should go into this directory.
[...]

PR #810, opened 2024-03-25
Revision 0.32.20

sinclairzx81:
This PR resolves a TypeScript compiler regression that occurred in between TS 5.3.3 -> 5.4.2. The issue seems isolated to the `value/delta/...` modules where the inferred TSchema types were emitting with basePath relative `src/...` paths, and not the relative module path. 

```typescript
// TS 5.3.3 - correct
export type Update = Static<typeof Update>;
export declare const Update: import("../../type/object/object.mjs").TObject<{
    type: import("../../type/literal/literal.mjs").TLiteral<"update">;
    path: import("../../type/string/string.mjs").TString;
    value: import("../../type/unknown/unknown.mjs").TUnknown;
}>;

// TS 5.4.2 - incorrect
export type Update = Static<typeof Update>;
export declare const Update: import("src/type/object/object.mjs").TObject<{
    type: import("src/type/literal/literal.mjs").TLiteral<"update">;
    path: import("src/type/string/string.mjs").TString;
    value: import("src/type/unknown/unknown.mjs").TUnknown;
}>;
```
This PR resolves the problem by explicitly importing the types using the correct path. Comments have been left to review the compiler behavior in subsequent revisions. It may also be worth introducing post build type assertions on top of the ESM module asserts (for consideration), but would prefer to trust compiler emit.

Fixes https://github.com/sinclairzx81/typebox/issues/808

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
         "ajv-formats": "^2.1.1",
         "mocha": "^9.2.2",
         "prettier": "^2.7.1",
-        "typescript": "^5.4.2"
+        "typescript": "^5.4.3"
       }
     },
     "node_modules/@andrewbranch/untar.js": {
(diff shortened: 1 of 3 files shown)


mikicho:
Thanks for the quick response!!

PR #922, opened 2024-07-01
Pin TypeScript to minor version since it is not semantically versioned

kevinlul:
See microsoft/TypeScript#14116

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
         "ajv-formats": "^2.1.1",
         "mocha": "^10.4.0",
         "prettier": "^2.7.1",
-        "typescript": "^5.5.2"
+        "typescript": "~5.5.3"
       }
     },
     "node_modules/@andrewbranch/untar.js": {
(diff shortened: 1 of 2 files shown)


sinclairzx81:
@kevinlul Hi, thanks for this PR

At this stage, I'm happy with the current setup as TypeBox pulls across (and tests) a variety of compiler versions and generally updates to the latest published version of the compiler via `@latest`. Making a distinction between semver or otherwise doesn't really implicate the project as it's generally built and updated to work with the latest published compiler while testing several exact [prior versions](https://github.com/sinclairzx81/typebox/blob/master/hammer.mjs#L35) (with much of the testing done via project tasks)

Thanks tho, PR's are appreciated, but will close this one out and things are fairly good here.
Cheers
S

kevinlul:
I should have checked https://github.com/sinclairzx81/typebox/blob/master/.github/workflows/nightly.yml, I missed that you're testing against typescript@latest and typescript@next so the actual version in devDependencies is immaterial.

PR #994, opened 2024-09-14
Avoid mutating if no union value matches

### LINKED ISSUE #993: `Value.Default` corrupts data in unions
In [FromUnion](https://github.com/sinclairzx81/typebox/blob/d049293b2302374e747297110f2f897d5b558cfb/src/value/default/default.ts#L145), it iterates over each of the union values, applies the defaults for that, and returns that defaulted type if it matches. However, it keeps mutating the same value, so this can get increasingly corrupted, leading to a false negative when the original value would have been valid. Example:

```typescript
const Tuple = Type.Tuple([Type.String(), Type.Number()])
const Array = Type.Array(Type.String())
const Union = Type.Union([Tuple, Array])

const withDefaults = Value.Default(Union, ['hello'])

console.log('withDefaults:', withDefaults)
console.log('errors:', [...Value.Errors(Union, withDefaults)])
```
outputs:
```
withDefaults: [ 'hello', undefined ]
errors: [
  {
    type: 62,
    schema: { anyOf: [Array], [Symbol(TypeBox.Kind)]: 'Union' },
    path: '',
    value: [ 'hello', undefined ],
    message: 'Expected union value'
  }
]
```
The default value for the tuple type pushes an `undefined` into the array, which doesn't match the tuple type. Next, it's checked against the `string[]` type, which doesn't match because of the `undefined`. Similar issues could arise with object properties.

---
As an aside, I question whether the default for tuples should ever change the length of an array. A tuple of one length is incompatible with the type of a tuple of a different length.
```typescript
type Tuple3 = [string, string, string]

const value: Tuple3 = ['a', 'b', 'c']

// ERROR: typescript: Tuple type 'Tuple3' of length '3' has no element at index '4'. [2493]
console.log(value[4])
value[4] = 'e'

// typescript: Type '"e"' is not assignable to type 'undefined'. [2322]
value[4] = 'e'
```
JavaScript just happens to implement them as arrays, but to the type system, the type is more like:
```typescript
type Tuple3 = [string, string, string, ...undefined[]]
```

Thus, I would expect the behavior to be:
```typescript
const Tuple = Type.Tuple([Type.String(), Type.Number({ default: 123 })])

Value.Default(Tuple, ['hello', undefined]) // ['hello', 123]
// not a valid value when we type check
Value.Default(Tuple, ['hello']) // ['hello']
```
But the above seems particularly incorrect when the tuple member that's "missing" doesn't even have a `default` property.

ehaynes99:
Fixes #993

### DIFF
--- a/src/value/default/default.ts
+++ b/src/value/default/default.ts
@@ -142,7 +142,8 @@ function FromTuple(schema: TTuple, references: TSchema[], value: unknown): any {
 function FromUnion(schema: TUnion, references: TSchema[], value: unknown): any {
   const defaulted = ValueOrDefault(schema, value)
   for (const inner of schema.anyOf) {
-    const result = Visit(inner, references, defaulted)
+    const result = Visit(inner, references, Clone(defaulted))
+    
     if (Check(inner, result)) {
       return result
     }
--- a/test/runtime/value/default/union.ts
+++ b/test/runtime/value/default/union.ts
@@ -82,6 +82,15 @@ describe('value/default/Union', () => {
     const R = Value.Default(T, { x: 3, y: 4 })
     Assert.IsEqual(R, { x: 3, y: 4 })
   })
+  it('Should return the original value if no schemas match', async () => {
+    const T = Type.Union([
+      Type.Tuple([Type.Number(), Type.Number()]),
+      Type.Array(Type.Number()),
+    ])
+    const value = ['hello']
+    const R = Value.Default(T, value)
+    Assert.IsTrue(R === value)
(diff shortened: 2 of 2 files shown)


sinclairzx81:
@ehaynes99 Hi, Hey thanks for this PR! (this union issue was a good catch)

Everything looks good here. Will merge this one in and publish shortly :)

sinclairzx81:
@ehaynes99 Heya,

Updates have been published on 0.33.11 :) 

Thanks again!
S

PR #999, opened 2024-09-18
Avoid losing precision when converting to bigints

### LINKED ISSUE #998: Avoid losing precision when converting to bigints
Hey!

Currently, `tryConvertBigInt` uses `BigInt(parseInt(value))` (see [here](https://github.com/sinclairzx81/typebox/blob/master/src/value/convert/convert.ts#L122)). But the `parseInt` step loses precision for large numbers.
Eg. `parseInt("12345678901234567890")` gives `12345678901234567000`, whereas BigInt("12345678901234567890") works correctly.

`parseInt` is being used because the bigint constructor doesn't work if you pass in a fractional string, eg. `BigInt("123.123")` throws an error whereas `BigInt(parseInt("123.123"))` does not.

So in order to avoid losing precision but also avoid throwing errors for fractional strings, we could skip the `parseInt` step if we know for sure that the value isn't fractional.

We could use something like `/\.|,/i.test(value)` to check if the input string may be fractional. This simply returns true if the input contains a '.' or a ','. There may well be "nicer" ways of performing this check though!

hpeebles:
Closes #998

### DIFF
--- a/src/value/convert/convert.ts
+++ b/src/value/convert/convert.ts
@@ -67,6 +67,9 @@ import { IsOptional } from '../../type/guard/kind'
 function IsStringNumeric(value: unknown): value is string {
   return IsString(value) && !isNaN(value as any) && !isNaN(parseFloat(value))
 }
+function IsStringBigIntLike(value: unknown): value is string {
+  return IsString(value) && /^-?\d+$/.test(value)
+}
 function IsValueToString(value: unknown): value is { toString: () => string } {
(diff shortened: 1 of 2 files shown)


sinclairzx81:
@hpeebles Hey, thanks for the PR!

Everything looks good here. Also thanks for the additional two tests. Will merge this one through and let you know once published.

Cheers!
S

sinclairzx81:
@hpeebles Hiya! Updates published on 0.33.12. 

Thanks again for the PR, this one was a very good catch :)

All the best!
S

PR #1013, opened 2024-10-02
feat: attach to ValueError the first error of each Union.anyOf visit

aleclarson:
As discussed in #779.

### Notes

1. It avoids the array allocation for cases where the first type in the union is a match.

2. It avoids wrapping the inner ValueError if only one error was found (which happens only if the union consists of a single type).


sinclairzx81:
@aleclarson Hi, thanks for the PR :)

Ah, I have been meaning to get around to this.  I am a bit reluctant to bring on new functionality here under a general update, but your PR is definitely on the right track in terms of how union errors would eventually land. The planned design was to implement iterators for each union sub variant. This means that in order to obtain union errors, callers will need to explicitly request them via exterior iterator expansion.

In saying this, it may be possible to bring this functionality as a general update, but would need the following amendments. 

### ValueError 

Update ValueError interface to contain a required `ValueErrorIterator` array. This array will always be present on generated errors, but empty if there are no interior errors.

```typescript
export interface ValueError {
  type: ValueErrorType
  schema: TSchema
  path: string
  value: unknown
  message: string
  errors: ValueErrorIterator[]
}
```

### Error Create

Update the Create function to accept a `ValueErrorIterator[]`. Will also need to update the `GetErrorFunction` to accept the errors as a parameter (this way callers can map union variant errors)

```typescript
function Create(errorType: ValueErrorType, schema: TSchema, path: string, value: unknown, errors: ValueErrorIterator[] = []): ValueError {
  return { type: errorType, schema, path, value, errors, message: GetErrorFunction()({ errorType, path, schema, value, /** todo: add error here */ }) }
}
```

### From Union

Bring in `Check` to perform a fast check union / value check and return immediately if ok. If there is an error, can remap a each variant into a ValueErrorIterator. Some of these union variants will be empty, but we expect at least 1 of the variants to contain a error.

```typescript
import { Check } from '../value/check/check'

// ...

function* FromUnion(schema: TUnion, references: TSchema[], path: string, value: any): IterableIterator<ValueError> {
  // Perform fast check across union to determine if there is an error  
  if(Check(schema, references, value)) return
  // Generate a ValueErrorIterator per union variant.
  const errors: ValueErrorIterator[] = schema.anyOf.map(variant => new ValueErrorIterator(Visit(variant, references, path, value)))
  // Yield union error
  yield Create(ValueErrorType.Union, schema, path, value, errors)
}
```

### Usage

Here is the anticipated usage.

```typescript
// Type

const U = Type.Union([
  Type.Literal('A'),
  Type.Literal('B'),
  Type.Literal('C')
])

const T = Type.Object({
  value: U
})

// .errors is ValueErrorIterator[]

const E = Value.Errors(T, { value: 'D' }).First()! // const E = {
                                                   //   type: 62,
                                                   //   schema: {
                                                   //     anyOf: [ [Object], [Object], [Object] ],
                                                   //     [Symbol(TypeBox.Kind)]: 'Union'
                                                   //   },
                                                   //   path: '/value',
                                                   //   value: 'D',
                                                   //   errors: [ // <--- variant errors here
                                                   //     ValueErrorIterator { iterator: Object [Generator] {} },
                                                   //     ValueErrorIterator { iterator: Object [Generator] {} },
                                                   //     ValueErrorIterator { iterator: Object [Generator] {} }
                                                   //   ],
                                                   //   message: 'Expected union value'
                                                   // }

// Enumerating errors requires recursive expansion of the interior
// variant errors. This can be implemented manually, but TB may 
// provide an API for this on the ValueErrorIterator in future
// revisions.

type ReducedError = [{ message: string, path: string, errors: ReducedError[] }]

const MapValueError = (error: ValueError) => ({
  message: error.message, 
  path: error.path,
  errors: error.errors.map(iterator => 
    MapValueIterator(iterator)
  )
})

const MapValueIterator = (iterator: ValueErrorIterator): ReducedError[]  => (
  [...iterator].map(error => MapValueError(error)) as never
)



const R = MapValueIterator(Value.Errors(T, { value: { x: 1 } }))

console.dir(R, { depth: 100 }) // const R = [
                               //   {
                               //     message: 'Expected union value',
                               //     path: '/value',
                               //     errors: [
                               //       [ { message: "Expected 'A'", path: '/value', errors: [] } ],
                               //       [ { message: "Expected 'B'", path: '/value', errors: [] } ],
                               //       [ { message: "Expected 'C'", path: '/value', errors: [] } ]
                               //     ]
                               //   }
                               // ]
```

### Rational for Errors as Iterators 

An important aspect of the Errors submodule is that Errors are not generated up front, rather they are pulled iteratively by the caller. This design is meant to provide a mechanism to protect against DOS issues where if users submit invalid data for a complex Union, this could explode into millions of errors (enough to overwhelm a system). Iterators enable the caller to terminate without exhausting all errors. 

If implementing sub variant union errors, these will also need to be iterators. This would provide the option for users to expand into the variants (depth + 1). Additionally, there are plans to make Intersect and Object Property errors require a depth + 1 expansion (which would require a minor revision / breaking change), but I think implementing them specifically for Union today should be fine.

---

Happy to discuss the above more, but would you be able to make the above updates?

Let me know!
S

aleclarson:
Closed in favor of #1015, which implements much of your suggestion :)

PR #1014, opened 2024-10-03
chore: specify prettier settings for vscode

aleclarson:
- Enable “format on save” in vscode for typescript files.
- Set prettier options in `.vscode/settings.json` so the right formatting is used.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,5 +4,13 @@
         "package-lock.json": true
     },
     "editor.suggest.showStatusBar": false,
-    "typescript.tsdk": "node_modules\\typescript\\lib"
+    "typescript.tsdk": "node_modules\\typescript\\lib",
+    "prettier.semi": false,
+    "prettier.singleQuote": true,
+    "prettier.printWidth": 240,
+    "prettier.trailingComma": "all",
+    "[typescript]": {
+        "editor.formatOnSave": true,
+        "editor.defaultFormatter": "esbenp.prettier-vscode"
+    }
 }


aleclarson:
We might want to use a `.prettierrc` file instead of duplicating the Prettier config into `.vscode/settings.json`

sinclairzx81:
@aleclarson Hey, thanks for the PR on this one.

Unfortunately, I am probably going to hold off on this for the time being as I am actually considering moving away from prettier (most of the lines in this codebase have `prettier-ignore` due to the issues it has with formatting types). I do think the project needs a formatter, but not keen to invest too much more into automation around the prettier library (other than a manual `npm run format`)

I have actually been considering moving the TypeBox toolchain over to Deno which supports `fmt`, `lint` and `test` tooling. I think if that's been achieved, I would be interested in exploring project automation specifically around the Deno toolchains, but until such time, I'm quite keen to keep this aspect of the project as is.

Thanks again for the PR, would be happy to explore this down the road, but will close this out for now.
Cheers!
S

aleclarson:
I'd like you to reconsider this one.
1. It's not a complex change, so easy to remove (a simple `git revert` actually) once you migrate away from Prettier, which appears to be down your list of priorities anyway.
2. Running `npm run format` is annoying, because it has to run on every file, and TypeBox has... well, *a lot* of files.

sinclairzx81:
@aleclarson Hiya,

Um, let me think about this some more. I have a few reservations around coupling automation tasks to editors, but not entirely adverse to the idea of getting better automation up around the project. If it helps developers that's a good thing (but wouldn't want "format on save" to be a detriment to writing types internal to the library (where formatters have proven to be more painful than helpful)

I am working on some new components for TypeBox which is taking the priority right now. But will come back to this issue following that (as there is some cross over between these new components and conventions for formatting)

Will keep you posted!

PR #1015, opened 2024-10-03
feat: add `errors` property to ValueError

aleclarson:
Follow-up to #1013, inspired by [this comment](https://github.com/sinclairzx81/typebox/pull/1013#issuecomment-2390217176)

**Copilot explained this PR:**

This pull request introduces enhancements to the error handling system in the `src/errors/errors.ts` file and updates the corresponding tests in `test/runtime/errors/types/union.ts`. The main changes involve adding nested error details to the `ValueError` interface and improving the handling of union type errors.

### Enhancements to error handling:

* [`src/errors/errors.ts`](https://github.com/sinclairzx81/typebox/pull/1015/files#diff-0bae23b87c88508b9858e3b732df4e42040118a2b22ed6ecab4555503bd6e7a1R37): Added a new import for `Check` and updated the `ValueError` interface to include a new `errors` property, which is an array of `ValueErrorIterator` instances. [[1]](https://github.com/sinclairzx81/typebox/pull/1015/files#diff-0bae23b87c88508b9858e3b732df4e42040118a2b22ed6ecab4555503bd6e7a1R37) [[2]](https://github.com/sinclairzx81/typebox/pull/1015/files#diff-0bae23b87c88508b9858e3b732df4e42040118a2b22ed6ecab4555503bd6e7a1R171)
* [`src/errors/errors.ts`](https://github.com/sinclairzx81/typebox/pull/1015/files#diff-0bae23b87c88508b9858e3b732df4e42040118a2b22ed6ecab4555503bd6e7a1L208-R218): Modified the `Create` function to accept an optional `errors` parameter and include it in the returned `ValueError` object.
* [`src/errors/errors.ts`](https://github.com/sinclairzx81/typebox/pull/1015/files#diff-0bae23b87c88508b9858e3b732df4e42040118a2b22ed6ecab4555503bd6e7a1L519-R530): Improved the `FromUnion` function to utilize the `Check` function and yield detailed errors for each variant in the union type.

### Updates to tests:

* [`test/runtime/errors/types/union.ts`](https://github.com/sinclairzx81/typebox/pull/1015/files#diff-1f872a1cc079015a2cdb8a7414aaf317c8d6ea413650f38793cee37698ffe2e9L2-R2): Updated the import statements to include `ValueErrorIterator` and added assertions to check the nested error details for union type errors. [[1]](https://github.com/sinclairzx81/typebox/pull/1015/files#diff-1f872a1cc079015a2cdb8a7414aaf317c8d6ea413650f38793cee37698ffe2e9L2-R2) [[2]](https://github.com/sinclairzx81/typebox/pull/1015/files#diff-1f872a1cc079015a2cdb8a7414aaf317c8d6ea413650f38793cee37698ffe2e9R20-R21)


sinclairzx81:
@aleclarson Hiya, thanks for the updated PR

> - Add errors: ValueErrorIterator | null to ValueError interface.
>
> - It yields the error of every failed Visit for the union.

So, will probably want to go with the `errors: ValueErrorIterator[]` over the `errors: ValueErrorIterator | null`. The reason for this is to be able to differentiate what union variant generated what error. Keep in mind that it's possible for each sub variant to yield multiple errors (some of which may be duplicate errors spanning multiple variants), and applications may wish to associate the error against the particular variant (this wouldn't be possible if flattening all errors into a single iterator)

Also, keen for `errors: ValueErrorIterator[]` over `errors: ValueErrorIterator[] | null`, mostly because this makes enumeration of interior errors simpler without having to check for `null`. Checking for `errors.length > 0` would be the alternative if actually needing to check, but mostly implementations can use enumerate on the errors `[]` without any checks at all (which is a nice property of array), I am generally fine with paying a allocation cost for `[]`.

Here were the structures / usage from #1013 

```typescript
// Type

const U = Type.Union([
  Type.Literal('A'),
  Type.Literal('B'),
  Type.Literal('C')
])

const T = Type.Object({
  value: U
})

// .errors is ValueErrorIterator[]

const E = Value.Errors(T, { value: 'D' }).First()! // const E = {
                                                   //   type: 62,
                                                   //   schema: {
                                                   //     anyOf: [ [Object], [Object], [Object] ],
                                                   //     [Symbol(TypeBox.Kind)]: 'Union'
                                                   //   },
                                                   //   path: '/value',
                                                   //   value: 'D',
                                                   //   errors: [ // <--- variant errors here
                                                   //     ValueErrorIterator { iterator: Object [Generator] {} },
                                                   //     ValueErrorIterator { iterator: Object [Generator] {} },
                                                   //     ValueErrorIterator { iterator: Object [Generator] {} }
                                                   //   ],
                                                   //   message: 'Expected union value'
                                                   // }

// Enumerating errors requires recursive expansion of the interior
// variant errors. This can be implemented manually, but TB may 
// provide an API for this on the ValueErrorIterator in future
// revisions.

type ReducedError = [{ message: string, path: string, errors: ReducedError[] }]

const MapValueError = (error: ValueError) => ({
  message: error.message, 
  path: error.path,
  errors: error.errors.map(iterator => 
    MapValueIterator(iterator)
  )
})

const MapValueIterator = (iterator: ValueErrorIterator): ReducedError[]  => (
  [...iterator].map(error => MapValueError(error)) as never
)



const R = MapValueIterator(Value.Errors(T, { value: { x: 1 } }))

console.dir(R, { depth: 100 }) // const R = [
                               //   {
                               //     message: 'Expected union value',
                               //     path: '/value',
                               //     errors: [
                               //       [ { message: "Expected 'A'", path: '/value', errors: [] } ],
                               //       [ { message: "Expected 'B'", path: '/value', errors: [] } ],
                               //       [ { message: "Expected 'C'", path: '/value', errors: [] } ]
                               //     ]
                               //   }
                               // ]
```
Note that the result `R` in this implementation is expanding all iterator errors for each variant, it would be possible to update this logic to only pull the `.First()` error, which would result in the following structure.

```typescript
// const R = [
//   {
//     message: 'Expected union value',
//     path: '/value',
//     errors: [ 
//       { message: "Expected 'A'", path: '/value', errors: [] }, // ... if pulling only the first error from each
//       { message: "Expected 'B'", path: '/value', errors: [] }, //    sub variant
//       { message: "Expected 'C'", path: '/value', errors: [] }
//     ]
//   }
// ]
```
Having the flexibility to map errors in either case (First or Spread) is desirable, even if the downstream implementations are a bit complex. They key would be that it should be possible to write examples of mapping errors in different ways, TypeBox can look at implementing helper functions as a subsequent task / PR. 

Let me know your thoughts!
S

aleclarson:
Thanks for explaining. Now I get it :)

I've made the changes as requested.

sinclairzx81:
@aleclarson Awesome thanks! This is looking good :)

There is just one more quick thing that would need to be updated for the PR to merge. Just need to update the `ErrorFunctionParameter` type to include the `errors` as an additional parameter.

https://github.com/sinclairzx81/typebox/blob/master/src/errors/function.ts#L172-L181

```typescript
// ErrorFunction (note: could rename this comment also)
export type ErrorFunctionParameter = {
  /** The type of validation error */
  errorType: ValueErrorType
  /** The path of the error */
  path: string
  /** The schema associated with the error */
  schema: TSchema
  /** The value associated with the error */
  value: unknown
  /** (ADDED) Interior errors for this error */
  errors: ValueErrorIterator[]
}
```

This makes it possible for sub variant error messages to be generated via `SetErrorFunction`. We don't need to do anything implementation wise, just make the error iterators available for potential mapping.

I think once the ErrorFunctionParameter is updated, the PR should be good to merge

Good work!

aleclarson:
Done. I kept the comment as it was, since it appears to be describing more than just the `ErrorFunctionParameter` type.

aleclarson:
I believe I fixed the failed check. Should be good to merge now?

sinclairzx81:
@aleclarson Hey, awesome!

I've merged your PR and published it in version 0.33.16. Great work!

I'm quite keen on this PR. It really does open up some interesting possibilities for exploring sub-error iterators in other structures, such as object properties, array/tuple elements, intersections, and more. For Objects specifically, the top-level iterator could only yield an "Expected Object" error, with deeper error iterators revealing any additional issues on properties. I think expanding error handling via iterators is an good design direction, and a potential focus in future revisions.

Thanks again for your contribution, really appreciate it!

All the best!
S

PR #1019, opened 2024-10-03
Fix converting large numbers to BigInts

### LINKED ISSUE #1018: BigInt conversion fails for numbers greater than 2^31
Hey!

I've just run into an issue when converting numbers to bigints where some of them become negative during the `BigInt(value | 0)` step.

For example: `BigInt(3155399673 | 0)` returns -1139567623n.

The union with 0 is done to remove decimals, and it works for small numbers, but it turns out that this first converts the number to a 32bit int, which means numbers greater than 2^31 are incorrectly converted.

I can have a go at fixing this later.

hpeebles:
Fixes #1018

### DIFF
--- a/src/value/convert/convert.ts
+++ b/src/value/convert/convert.ts
@@ -119,9 +119,8 @@ function TryConvertBoolean(value: unknown) {
   return IsValueTrue(value) ? true : IsValueFalse(value) ? false : value
 }
 function TryConvertBigInt(value: unknown) {
-  value = IsNumber(value) ? value.toString() : value;
   const truncateInteger = (value: string) => value.split('.')[0]
-  return IsStringNumeric(value) ? BigInt(truncateInteger(value)) : IsValueFalse(value) ? BigInt(0) : IsValueTrue(value) ? BigInt(1) : value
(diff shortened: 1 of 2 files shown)


hpeebles:
The existing conversion from string to BigInt works nicely even when the number is not an integer, so this change make it such that numbers are first converted to strings before converting to BigInt.
Prior to this, the fractional component of any numbers was removed by doing `BigInt(value | 0)`, but this treats `value` as a 32 bit integer which means numbers larger than 2^31 are converted incorrectly.

hpeebles:
Another update... I've now switched to using `Math.trunc` which does exactly what we want for both positive and negative numbers, and works for large numbers (up to `Number.MAX_SAFE_INTEGER`).

See here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc.

sinclairzx81:
@hpeebles Hi, thanks for the PR! :)

Yeah, this looks good to me (and it's great to have someone looking at this side of the codebase). Will merge this one through, and publish shortly.

Thanks again!
S

sinclairzx81:
@hpeebles Hiya

Updates published on 0.33.14 

Thanks again!
S

PR #1021, opened 2024-10-04
Simplify module

y0x1:
It feels to me that the conditional could be simplified, but I am not sure if there might be any potential downsides. Especially problematic might be if the conditional has not been exhaustive.

Most likely the updated version should be a bit faster.

I checked the readme example and seems to work fine.


sinclairzx81:
@y0x1 Hi, Thanks for the PR,

I think you make a really good point about reducing the number of conditional arms to check (resolving on the precomputed / cached `static` would certainly help here) but I think I am keen to press on with the full on conditional matching (as there are plans to potentially ditch the `static` property entirely in an upcoming revision, and I would like to test inference extents on the non-optimized form). Also prototypes can be modified by end users as need be (they are copy / paste)

It would be interesting to compare performance between full conditional match and the cached type. You should be able to use the following package to get a bit of insight there.

https://www.npmjs.com/package/@arktype/attest

If you were to compare performance there and relay information back to me, that would be hugely helpful to the project. If the tests show a marginal performance increase on the cached type, I will investigate retaining `static` and we can go with this optimization then.

Will close of this PR for now, but can reopen post measurement. 
Cheers
S

PR #1026, opened 2024-10-08
Relax String Pattern matching

sirlancelot:
Allow newlines in string pattern when using `TemplateLiteral`.

### DIFF
--- a/src/type/patterns/patterns.ts
+++ b/src/type/patterns/patterns.ts
@@ -28,7 +28,7 @@ THE SOFTWARE.
 
 export const PatternBoolean = '(true|false)'
 export const PatternNumber = '(0|[1-9][0-9]*)'
-export const PatternString = '(.*)'
+export const PatternString = '([\S\s]*)'
 export const PatternNever = '(?!.*)'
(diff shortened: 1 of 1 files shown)


sinclairzx81:
@sirlancelot Hi, thanks for the PR.

Unfortunately, I can't accept this PR as the `PatternString` is used in quite a few places (for example Record keys), and is not just limited to just TemplateLiteral. This change actually breaks 28 tests spanning Record, TemplateLiteral and Intrinsic string manipulation types, so cannot be merged.

Overall, there would need to be careful consideration of this change as well as a suite of associated tests written to ensure multiline strings do not break assumptions made elsewhere throughout the codebase. In the case of TemplateLiteral, tests would also need to extend to static inference (where multiline strings may require additional inference paths written)

Will close this one up for now. Just note that it's best to submit an issue before submitting a PR. As per contribution notice:

https://github.com/sinclairzx81/typebox?tab=readme-ov-file#contribute

> TypeBox is open to community contribution. Please ensure you submit an open issue before submitting your pull request. The TypeBox project prefers open community discussion before accepting new features.

Cheers
S

PR #1030, opened 2024-10-11
Fix optional array schema converting `undefined` to `[undefined]`

### LINKED ISSUE #1029: Converting `undefined` using an optional array schema returns `[undefined]`
Hey!

I've just run into an issue where I have a schema of the form `Type.Optional(Type.Array(...))`, if I call `Value.Convert` using this schema and pass in a value of `undefined` it returns `[undefined]` whereas it should return `undefined`.

Here is a minimal failing test -

```
it('Should not convert optional array if input is undefined', () => {
  const T = Type.Optional(Type.Array(Type.Number()));
  const R = Value.Convert(T, undefined); // returns [undefined]
  Assert.IsEqual(R, undefined)
})
```

hpeebles:
Fixes #1029

### DIFF
--- a/src/value/convert/convert.ts
+++ b/src/value/convert/convert.ts
@@ -170,6 +170,7 @@ function Default(value: unknown): unknown {
 // Convert
 // ------------------------------------------------------------------
 function FromArray(schema: TArray, references: TSchema[], value: any): any {
+  if (value === undefined) return undefined;
   const elements = IsArray(value) ? value : [value]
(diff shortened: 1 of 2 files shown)


sinclairzx81:
@hpeebles Heya,

Will close this one off inline with the issue for now. But happy to discuss additional aspects of Convert over on https://github.com/sinclairzx81/typebox/issues/1029 

Cheers!
S

PR #1052, opened 2024-10-31
feat: allow importing Decode/Encode directly

### LINKED ISSUE #876: How to import just Decode?
```js
import { Value } from '@sinclair/typebox/value';
 const appSettings = Value.Decode(AppSettings, body);
```
Is there a way to just use Decode for better tree shaking?

I tried importing like this but it gave build error:
```js
import { Decode } from '@sinclair/typebox/build/cjs/value/value/value';
```

aleclarson:
Fixes #876

ESBuild *has no support* for tree-shaking of **re-exported namespaces**, which is what TypeBox is doing with its `Value` namespace.

By separating `Decode` and `Encode` into standalone modules, ESBuild can shake anything else not used in the `@sinclair/typebox/value` module.

### DIFF
--- a/src/value/decode/decode.ts
+++ b/src/value/decode/decode.ts
@@ -26,35 +26,20 @@ THE SOFTWARE.
 
 ---------------------------------------------------------------------------*/
 
-// ------------------------------------------------------------------
-// Errors (re-export)
-// ------------------------------------------------------------------
-export { ValueError, ValueErrorType, ValueErrorIterator } from '../errors/index'
-// ------------------------------------------------------------------
-// Guards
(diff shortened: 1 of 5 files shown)


sinclairzx81:
@aleclarson Hi, apologies for the delay here.

I think this PR looks good and I really quite like the refactoring here (removing the duplication of interfaces on `value.ts` is a nice win). I'll merge this one through and carry out a couple of tests, will publish out on the next revision. 

Again, sorry for the delay.
Many thanks!
S

sinclairzx81:
@aleclarson Hiya,

Updates published on 0.33.18

Many thanks!
S

PR #1076, opened 2024-11-15
Fix records in modules

zoriya:
Without this, all record in modules were inferred as `never`.

### DIFF
--- a/src/type/module/module.ts
+++ b/src/type/module/module.ts
@@ -26,7 +26,7 @@ THE SOFTWARE.
 
 ---------------------------------------------------------------------------*/
 
-import { Ensure } from '../helpers/index'
+import { Assert, Ensure } from '../helpers/index'
 import { CreateType } from '../create/index'
 import { Kind } from '../symbols/index'
 import { SchemaOptions, TSchema } from '../schema/index'
@@ -58,6 +58,7 @@ import { TUndefined } from '../undefined/index'
 import { TUnknown } from '../unknown/index'
 import { TVoid } from '../void/index'
 import { Static } from '../static/index'
+import { TRecord } from '../record/index'
 
 // ------------------------------------------------------------------
 // Infer
@@ -73,6 +74,10 @@ type InferRef<Ref extends string, Properties extends TProperties> = (
 // prettier-ignore
 type InferObject<Properties extends TProperties, ModuleProperties extends TProperties> = { 
   [K in keyof Properties]: Infer<Properties[K], ModuleProperties> 
+} & {}
+// prettier-ignore
+type InferRecord<Properties extends TSchema, S extends TSchema, ModuleProperties extends TProperties> = { 
+  [_ in Assert<Static<Properties>, PropertyKey>]: Infer<S, ModuleProperties> 
 } & {}
 // prettier-ignore
 type InferConstructor<Parameters extends TSchema[], InstanceType extends TSchema, Properties extends TProperties> = Ensure<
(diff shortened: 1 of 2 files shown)


sinclairzx81:
@zoriya Hi, thanks for this PR (and great work navigating this aspect of the code!)

At the moment, there is actually quite a lot of Module related work being done in the `ref` branch. Would it be possible to retarget your PR for `ref` instead of `master` ? Just note that `ref` branch moves the infer logic from `module.ts`, to `src/type/module/infer.ts`. This code will be going out on 0.44.1.

---

A lot of the work on the `ref` branch relates to mutual recursive / auto deref / computed types, this to enable the following (which isn't possible on 0.44.0)

```typescript
const Module = Type.Module({
   T: Type.Object({ x: Type.Number() }),
   B: Type.Partial(Type.Ref('T')) // where interior Ref is dereferenced via Import
})
```
Record actually fits into the computed type category and is yet to be worked on, so this PR would certainly help make a start with the inference side of things. Actually any help would be appreciated with the new Module feature! :)

Thanks again, Keep me posted.
S

sinclairzx81:
@zoriya Hey, just an update here.

It might be good to hold off retargeting to `ref` for the time being. I'm currently in the process of working through some fairly difficult challenges supporting deferred computed types, and things are a bit influx atm. There is a high likelihood of things changing quite a bit internally (inclusive of potentially implementing functions for each of the `TInferX` types (which would include Record also)). Some of the code around this is very intricate and spans a lot of library, so I'll need to be across most of it.

- Updates are relevant to the following computed types:
  - Awaited
  - Pick
  - Omit
  - Partial
  - Required
  - KeyOf
  - Record
  - Mapped
  - Extends
  - Extract
  - Exclude
  - Intrinsics (Uppercase, Lowercase)
  
I'll keep this PR here for the time being, but will need to defer merging through while I work through some of internals. 
Will keep you posted
S

sinclairzx81:
@zoriya Hiya!

Ended up having to implement some fairly sophisticated mapping for Record and other computable types. There is still quite a lot to do, but the implementation turned out to be less than trivial :( Can take a look at the updates below if you're interested.

https://github.com/sinclairzx81/typebox/blob/master/src/type/module/compute.ts#L189-L220

```typescript
// Computed
// prettier-ignore
type TFromComputed<ModuleProperties extends TProperties, Target extends string, Parameters extends TSchema[],
  Dereferenced extends TSchema[] = TDerefParameters<ModuleProperties, Parameters>
> = (
  Target extends 'Awaited' ? TFromAwaited<Dereferenced> :
  Target extends 'Index' ? TFromIndex<Dereferenced> :
  Target extends 'KeyOf' ? TFromKeyOf<Dereferenced> :
  Target extends 'Partial' ? TFromPartial<Dereferenced> :
  Target extends 'Omit' ? TFromOmit<Dereferenced> :
  Target extends 'Pick' ? TFromPick<Dereferenced> :
  Target extends 'Record' ? TFromRecord<Dereferenced> :
  Target extends 'Required' ? TFromRequired<Dereferenced> :
  TNever
)
// prettier-ignore
function FromComputed<ModuleProperties extends TProperties, Target extends string, Parameters extends TSchema[]>(moduleProperties: ModuleProperties, target: Target, parameters: Parameters): TFromComputed<ModuleProperties, Target, Parameters> {
  const dereferenced = DerefParameters(moduleProperties, parameters)
  return (
    target === 'Awaited' ? FromAwaited(dereferenced) :
    target === 'Index' ? FromIndex(dereferenced) :
    target === 'KeyOf' ? FromKeyOf(dereferenced) :
    target === 'Partial' ? FromPartial(dereferenced) :
    target === 'Omit' ? FromOmit(dereferenced) :
    target === 'Pick' ? FromPick(dereferenced) :
    target === 'Record' ? FromRecord(dereferenced) :
    target === 'Required' ? FromRequired(dereferenced) :
    Never()
  ) as never
}
```

I really appreciate the PR tho (and the time taken to go through the type mapping code). It's actually good to have people dig into the type mapping and submit fixes (there's a lot of it!). Will close out this one for now though as the Record got implemented via the new Computed Type infrastructure.

Thanks again, All the best!
S

zoriya:
Thanks for keeping me updated!

Types looked surprisingly readable for the typescript shenanigans that were done, typebox is really cool!

PR #1096, opened 2024-11-22
Fix importing optional properties

michaelcollabai:
Module types look like a great idea to simplify some of our recursive schemas.

I tried them out and noticed a problem with optional object properties:
When importing an object schema, the optional status is lost for all properties with the more complex types handled in `TFromType`.
This does not happen at run time as `[OptionalKind]` is carried over in the spread op in `CreateType`.
I added type and run time tests to verify my assumptions.

Hats off for building something so complex with such clean and readable code!

### DIFF
--- a/src/type/module/compute.ts
+++ b/src/type/module/compute.ts
@@ -364,7 +364,7 @@ export type TFromType<ModuleProperties extends TProperties, Type extends TSchema
 // prettier-ignore
 export function FromType<ModuleProperties extends TProperties, Type extends TSchema>(moduleProperties: ModuleProperties, type: Type): TFromType<ModuleProperties, Type> {
   return (
-    // Modifier Unwrap
+    // Modifier Unwrap - Reapplied via CreateType Options
     KindGuard.IsOptional(type) ? CreateType(FromType(moduleProperties, Discard(type, [OptionalKind]) as TSchema) as never, type) :
     KindGuard.IsReadonly(type) ? CreateType(FromType(moduleProperties, Discard(type, [ReadonlyKind]) as TSchema) as never, type) :
     // Traveral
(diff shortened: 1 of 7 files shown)


michaelcollabai:
I think TReadOnly works in a similar way so it probably has the same issue.
But I also see that the area is still under active development. Let me know if a fix is helpful.

sinclairzx81:
@michaelcollabai Hi!

Ah! Yes, this is a totally a bug (thanks for the follow up + repro!). This fix will need to be applied to both Optional and Readonly. You will want to update both `FromType` and `TFromType` just to keep these mappings symmetric (even if the FromType is just a pass through of the type, ill be coming back to this at a later time)

```typescript
// ...
Type extends TOptional<infer Type extends TSchema> ? TOptional<TFromType<ModuleProperties, Type>> :
Type extends TReadonly<infer Type extends TSchema> ? TReadonly<TFromType<ModuleProperties, Type>> :
```

Can you update for Readonly also, and to just drop a test in there for the following structure

```typescript
const Module = Type.Module({
  // will want to ensure the unwrap works for Readonly and Optional applied modifiers
  x: Type.ReadonlyOptional(Type.Array(Type.Literal(1))), 
  y: Type.Readonly(Type.Array(Type.Literal(1))),
  z: Type.Optional(Type.Array(Type.Literal(1))),
  w: Type.Array(Type.Literal(1))
})
```
If you can get the Readonly mapping in + this test, this PR should be good to go.
Thanks again, this was good spotting.
S

sinclairzx81:
@michaelcollabai Hi,

Have gone ahead and fixed up Readonly on this PR and merged through (this issue was a breaking bug and wanted to get it sorted pretty quickly). Will be publishing the fix out to NPM push within the hour. Thanks again for finding this issue and for the PR to resolve, this was really good catch (and happy to hear the code wasn't to bad to navigate!)

Will notify once published
Cheers!
S

sinclairzx81:
@michaelcollabai Hiya,

Fix published on 0.34.8. Thanks again!

All the best
S

PR #1149, opened 2025-01-30
Add functionality to convert custom types 

lecafard:
Add some functionality to convert custom types, as well as allowing data to be injected during the conversion process.

Also reworked the registry into classes to reduce duplication, keeping in mind backwards compatibility.


sinclairzx81:
@lecafard Hi, thanks for the PR. This one looks interesting (and sorry for the delay)

I'll need some time to sit down and properly review these updates (I am bit pressed with other projects atm), but did have a quick question regarding the rationale behind the additional `data` property to the Convert signatures.

I guess my initial thoughts is that metadata should be encoded and received via Schema only, but curious to know the thinking there.

---

The registry updates look good here (that's some great thinking). I'll need a few extra days to properly sit down and go through these updates, but at a high level, these seem good. Just let me know your thoughts on the additional `data` parameter, and can pick things up in a few days.

Cheers
S

lecafard:
Hi there, I originally added the data parameter because I had a use case where I needed to inject specific data during the conversion process (such as an encryption key) that would be unique to the object being converted.

If there's a better approach, please let me know. I understand this is a rather niche use case, which might not be a good fit for this library.

sinclairzx81:
@lecafard Hi, sorry for the delay in this PR

I think I am going to close this one out. I think in the immediate term, conversions could probably be handled via the TypeBox Transform (Encode/Decode), and a bit reluctant to introduce a specialized registry solely for Convert. I think having a good degree of programmatic control over each TypeBox function is a good idea (and Registries certainly help in this regard), but I think a more generalized solution that could apply to everything (Clean, Convert, Default, Cast, Create, etc) would be the better goal to strive for (although I don't know what form that would take, I think the idea is a good one)

Will close out for now, but happy to discuss more if you would like. 
Again, sorry for the extended delay on this.
All the best
S

PR #1194, opened 2025-03-04
feat: add entrypoints for guard modules

aleclarson:
Currently, there is no way to import `TypeGuard`, `KindGuard`, or `ValueGuard` in a way that allows for tree-shaking. So, even if you use one function on `TypeGuard`, every function is included in your bundle, which is not ideal for serverless environments.

With this PR, you can import like so:
```ts
import * as TypeGuard from '@sinclair/typebox/guard/type'
import * as KindGuard from '@sinclair/typebox/guard/kind'
import * as ValueGuard from '@sinclair/typebox/guard/value'
```


src/value/guard/index.ts
```diff
@@ -1,29 +0,0 @@
...
-IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-THE SOFTWARE.
-
----------------------------------------------------------------------------*/
-
-export * from './guard'
```
aleclarson, line 29: Completely removed `src/value/guard`, as I noticed some ValueGuard utilities were duplicated.

src/type/index.ts
```diff
@@ -46,7 +46,6 @@ export * from './exclude/index'
 export * from './extends/index'
 export * from './extract/index'
 export * from './function/index'
-export * from './guard/index'
```
aleclarson, line 49: May want to add this back as `../guard/type/index` so this isn't a breaking change.

src/syntax/syntax.ts
```diff
@@ -44,7 +45,7 @@ export function NoInfer(...args: any[]): t.TSchema {
   const withContext = typeof args[0] === 'string' ? false : true
   const [context, code, options] = withContext ? [args[0], args[1], args[2] || {}] : [{}, args[0], args[1] || {}]
   const result = Module.Parse('Type', code, context)[0]
-  return t.KindGuard.IsSchema(result)
```
aleclarson, line 47: Updated the `syntax` module to use namespace imports.

src/guard/value/index.ts
```diff
@@ -1,6 +1,6 @@
 /*--------------------------------------------------------------------------
 
-@sinclair/typebox/value
+@sinclair/typebox/type
```
aleclarson, line 3: Oops, need to update these headers.

sinclairzx81:
@aleclarson Hi, Sorry for the extended delay on this PR. 

Unfortunately I don't think I can accept this one as I am currently freezing library layouts in preparation for a very large future update to TypeBox (the next big revision push for `0.35.x`)

As of writing, TypeBox is currently in the process undergoing of a full redesign / rearchitecture to consolidate several years of functionality (as well as introducing a suite of new functionality). As part of that process, Static Types, Factory Types, Guards and Option Omitters (new) are all going to be merged into a single module per type. For example, here is the new layout for TString (inclusive of its TypeGuard)

```typescript
import { Guard } from '../../guard/index.ts'
import { Memory } from '../../system/index.ts'
import { type TSchema, type TOptions } from './schema.ts'

// Static
export type StaticString = string

// Type
export type TFormat = 
  | 'base64'
  | 'date-time'
  | 'date'
  | 'duration'
  | 'email'
  | 'hostname'
  | 'idn-email'
  | 'idn-hostname'
  | 'ipv4'
  | 'ipv6'
  | 'iri-reference'
  | 'iri'
  | 'json-pointer-uri-fragment'
  | 'json-pointer'
  | 'json-string'
  | 'regex'
  | 'relative-json-pointer'
  | 'time'
  | 'uri-reference'
  | 'uri-template'
  | 'url'
  | 'uuid'
  | ({} & string)

export interface TStringOptions extends TOptions {
  format?: TFormat
  minLength?: number
  maxLength?: number
  pattern?: string | RegExp
}

export interface TString extends TSchema {
  type: 'string'
}
// Factory
export function String(options?: TStringOptions): TString {
  return Memory.Create({}, { type: 'string' }, options) as never
}
// Guard
export function IsString(value: unknown): value is TString {
  return Guard.IsObject(value) 
    && Guard.HasPropertyKey(value, 'type') 
    && Guard.IsEqual(value.type, 'string')
}
// Options
export function StringOptions(type: TString): TOptions {
  return Memory.Discard(type, ['type'])
}
```

---

Notable Changes Related to Guards and Type Handling

- Remove `Kind`, `OptionalKind`, `ReadonlyKind` and other symbols from TypeBox entirely. Type differentiation will be based entirely upon structural checking, which means `TypeGuard` and `KindGuard` will be dropped in their current form (although `TypeGuard` may remain as a top level which just imports the `Is*` function from each type)
- The current `ValueGuard` (living in the `type` submodule) will be renamed to `Guard`. This will service as a common JavaScript value assertion module which will be used for Type, Value and Compiler submodules.
- The current `Type.Clone`, `Type.Create` are being moved to a new `Memory` module (now part of `system`) which will handle instancing, cloning and updating types. The `Memory` module is being used also to create hidden `{ enumerable: false }` discriminator properties which will replace the existing symbols.
- Attention given to tree-shaking, and general reduced bundle sizes. (I am currently trying to promote the new `Syntax` functionality as standard (and trying to achieve this on a 16KB maximum minified + gzip compression budget))
- And many many many other things 

The work going on currently is an attempt consolidate several years of updates into a more cohesive and lightweight system. However to permit this work to go on I am keen to keep the current project structure as is (as I am familiar with it) as it means less guess work for me trying to understand how things are cross linked (and so I can track and document changes for future upgrade guides), But mostly because anything changed now will ultimately be replaced in future.

---

Again, sorry for the delay in getting to this, but hope the above makes things a bit clearer with regards to some of the work going on behind the scenes. The work being done is almost certainly going to take several months to finish, and as such TypeBox is mostly in a frozen state until things are ready to bring over. I have made few exceptions for upstream functionality that can be merged through early (optimized syntax parsing got merged yesterday), but outside of optimizations and bug fixes, `0.34.x` is effectively locked down till next release.

Will close out this PR for now but happy to discuss more on some of these planned changes. I think given the depth of what's being considered for the next release, establishing good communication with users on regarding design thinking would be good.

Thanks again, and sorry for the delay.
Cheers
S

PR #1196, opened 2025-03-06
feat: add the list of union type members in union type errors

tinkermakar:
Praise: This lib is so great, I define a type once and use it both to validate the incoming API calls and confidently process these calls knowing that the type I rely on matches exactly the actual data slipping through validation. Many many thanks

The problem: The only issue I have so far with the usage of this lib is that it does not detail the members of a failed Union type validation. This is not informative enough for my FE peers to fix their payload just by looking at the returned validation error messages.

The solution: The relevant schema is already available in the scope of `DefaultErrorFunction`, all we need to do is to add the types on top of the generic `Expected union value` message, and that's what I do here.

Tests: For some weird reason, tests are failing if I run them in VSCode's `JavaScript Debug Terminal`. It makes it difficult to add new tests. However, in a regular terminal no test fails, so I think we're good.

Thank you.

### DIFF
--- a/src/errors/function.ts
+++ b/src/errors/function.ts
@@ -156,7 +156,9 @@ export function DefaultErrorFunction(error: ErrorFunctionParameter) {
     case ValueErrorType.Undefined:
       return 'Expected undefined'
     case ValueErrorType.Union:
-      return 'Expected union value'
+      const unionMembers = error.schema?.anyOf?.map((el: {type: string}) => el.type);
+      const unionMembersSuffix = unionMembers?.length ? ` (${unionMembers?.join(' | ')})` : '';
+      return 'Expected union value' + unionMembersSuffix;
     case ValueErrorType.Void:
       return 'Expected void'
     case ValueErrorType.Kind:


sinclairzx81:
@tinkermakar Hi, thanks for the PR, it's appreciated.

Unfortunately, the error generator in TypeBox is currently locked down in preparation for some future updates related to union error generation (so can't take on new functionality at this stage). This said, it is possible to configure TypeBox to generate errors inline with this PR.

```typescript
import { SetErrorFunction, DefaultErrorFunction, ValueErrorType } from '@sinclair/typebox/errors'

// Configure: Error Intercept
SetErrorFunction((parameter) => {
  // Detect union and apply error message logic
  if(parameter.errorType === ValueErrorType.Union) {
    const unionMembers = parameter.schema?.anyOf?.map((el: {type: string}) => el.type)
    const unionMembersSuffix = unionMembers?.length ? ` (${unionMembers?.join(' | ')})` : ''
    return 'Expected union value' + unionMembersSuffix
  } 
  // Otherwise, return as default
  return DefaultErrorFunction(parameter)
})

// Test
const T = Type.Union([
  Type.Object({ type: Type.Literal('A') }),
  Type.Array(Type.Number()),
  Type.Number(),
  Type.String(),
  Type.Boolean(),
  Type.Null()
])

const E = [...Value.Errors(T, undefined)]

console.log(E[0].message) 
// ^
// Expected union value (object | array | number | string | boolean | null)
```

---

Quite a few users have reported challenges getting good human readable error messages generated for Union. It's actually been a bit of a challenge for TypeBox and there's not really a clear cut way to handle them. 

The future updates being planned intend to de-linearize (tree-ify) the TypeBox errors and produce a tree like structure for composited types (like Union). In the case of Union, the generator will emit a top level `Expected Union` Error (as is current), but contain an interior sequence of iterators to yield errors for each union variant. This is actually partially implemented on the current version (Union only), but the plan is to move all type errors to this tree error structure (applicable for Union, Intersect, Array, Tuple, Record and Object). 

### Alternative Implementation Pulling from Error Iterators

In terms of error message generation, your code is on the right track, but TypeBox needs to avoid internal introspection for union variants (so can't generate sub error messages as part of the union error message). Ideally it needs to work towards some mappable structure (tree) where sub errors can be traversed and mapped via the SetErrorFunction. The reason for this mostly relates to i18n (internationalization) support where single type error messages can be translated easily into different languages, but can also be reduced into projected strings as (per your PR). The key is that the application has control over how the error message is projected.

The following shows an alternative implementation using iterators to pull sub variant Union errors.

```typescript
SetErrorFunction((parameter) => {
  if(parameter.errorType === ValueErrorType.Union) {
    const messages = [...parameter.errors].reduce((result, error) => {
      const interior = [...error].map(error => ValueErrorType[error.type])
      return [...result, ...interior]
    }, [] as string[])
    const unionMembersSuffix = messages.length ? ` (${messages.join(' | ')})` : ''
    return 'Expected union value' + unionMembersSuffix
  }
  return DefaultErrorFunction(parameter)
})
```

> Mapping error messages in this way is a WIP and there needs to be design work done to make the API simpler to use, but the above should give you some insight into how TypeBox may handle error message generation in future.

---

Hope this brings some insight into some of the thinking around TypeBox error generation. Will close off the PR for now, but happy to field any additional questions you have around error message generation.

All the best
S

tinkermakar:
Understood, many thanks for the detailed response

Issue #1219, opened 2025-04-02
replaceInlineImportSpecifiers should only affect relative imports?


jonasi:
The latest build (0.34.32), has an import in the esm build that looks like this

```
// @sinclair/typebox/build/esm/syntax/syntax.d.mts
import * as t from '@sinclair/typebox.mjs';
```

This is due to changes to `src/syntax/syntax.ts` in #1218 

```
- import * as t from '../type/index'
+ import * as t from '@sinclair/typebox'
```

sinclairzx81:
@jonasi Hi, thanks for reporting! (and so quickly too, that's super appreciated!)

I have just pushed a fix for this on 0.34.33, if you run into any issues, let me know.

---

### Cause

Most of the new parsing infrastructure was developed remotely (with TypeBox import there via `@sinclair/typebox`). Unfortunately, when integrating these components back into TypeBox, my tooling doesn't flag issues related to TypeBox trying to import itself. The specific issue (post-fix with .mjs) was actually caused by my current build tooling manually rewriting import specifiers for CJS and ESM dual publishing.

Looks like I might need some extra checks to catch cases like this.

---

### Build Tooling

I'm actually considering moving the TypeBox codebase over to Deno. I’ve been exploring tooling options there, and it seems pretty compelling. Cross-platform builds feel a lot easier in Deno compared to Node, mostly because Deno enforces `.ts` for specifiers, making mapping more predictable than Node’s somewhat loose approach (especially for CJS). Some initial work is already underway, but the shift probably won’t happen until the second half of this year.

---

Thanks again for reporting!
S

jonasi:
Appreciate the prompt fix! Thank you.

PR #1220, opened 2025-04-02
Revision 0.34.33

sinclairzx81:
This PR is a hotfix for 0.34.32 where an incorrect import specifier was used to import the types sub module.

Fixes: https://github.com/sinclairzx81/typebox/issues/1219


jonasi:
thank you!

Issue #1245, opened 2025-05-03
support special characters in regex pattern generation


didierdemoniere:
hi @sinclairzx81,

2 weeks ago i submitted a small PR to fix a bug that occur when regex is used to check object property names with special characters here: 

https://github.com/sinclairzx81/typebox/pull/1231

I just read the contribute section for the project and It now make sense why it's still there.
Relevant informations about this bug are in the PR, and the fix is obviously a proposition.
How do you think we should proceed from here ?

sinclairzx81:
@didierdemoniere Hi, Sorry for the delay.

> 2 weeks ago i submitted a small PR to fix a bug that occur when regex is used to check object property names with special characters here:

Again, sorry about that. I have been out of action for past few weeks and haven't really been in a position where I've been able to sit down and review PR's. I did take a quick look at the one you submitted and it looks ok, I just haven't had the time / resources available to merge it through.

> I just read the contribute section for the project and It now make sense why it's still there.
Relevant informations about this bug are in the PR, and the fix is obviously a proposition.

Yeah, the general preference for TypeBox is to submit an issue first. I should probably set up a template to make this more clear. This said, immediate PR's are generally welcome if they fixing issues, and the property escape issue looks like it fits that criteria, so it's all good.

---

Will look to merge this one through today.
Cheers
S

sinclairzx81:
@didierdemoniere Hiya,

> Relevant informations about this bug are in the PR, and the fix is obviously a proposition. How do you think we should proceed from here ?

Sorry, I went ahead and merged through, but as noted on the PR, the updates introduced breaking issues for Intersect Check functions (TypeCompiler / Value.Check) so I have had to revert. I had anticipated there was adequate testing applied to the PR and the updates looked harmless enough, but unfortunately, they were breaking Intersect tests, so can't be merged through.

Looking at the [original reproduction code](https://github.com/sinclairzx81/typebox/pull/1231#issue-3004939705), could this not be simplified to the following?

```typescript
const schema = t.Recursive((This) => t.Object({ 
  $or: t.Optional(t.Array(This)),
  $not: t.Optional(This),
  $like: t.Optional(t.String()) 
}));

console.dir(Value.Errors(schema, { 
  $like: 'foo', 
  $not: { $like: 'bar' } 
}).First(), { depth: 1000 })
```
Is there a reason you needed nested Recursive / Intersected schematics?

didierdemoniere:
hello again @sinclairzx81,

[...]

PR #1249, opened 2025-05-07
Fix: escape properties in regexp pattern generation 2

didierdemoniere:
resurection of https://github.com/sinclairzx81/typebox/pull/1231

issue: https://github.com/sinclairzx81/typebox/issues/1245

simpler reproduction:

```ts

const A = Type.Object({ $x: Type.Number() })
const B = Type.Object({ $y: Type.Number() })
const T = Type.Intersect([A, B], { unevaluatedProperties: false })

// should be undefined
Value.Errors(T, { $x: 1, $y: 1 }).First()
```

### DIFF
--- a/src/type/keyof/keyof-property-keys.ts
+++ b/src/type/keyof/keyof-property-keys.ts
@@ -50,9 +50,9 @@ type TFromRest<Types extends TSchema[], Result extends PropertyKey[][] = []> = (
     : Result
 )
 // prettier-ignore
-function FromRest<Types extends TSchema[]>(types: [...Types]): TFromRest<Types> {
+function FromRest<Types extends TSchema[]>(types: [...Types], options?: KeyOfPropertyKeysOptions): TFromRest<Types> {
   const result = [] as PropertyKey[][]
-  for(const L of types) result.push(KeyOfPropertyKeys(L))
+  for(const L of types) result.push(KeyOfPropertyKeys(L, options))
   return result as never
 }
 // ------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


sinclairzx81:
@didierdemoniere Hi, sorry for the extended delay on this.

I think this PR is fine in the regard in that it escapes for non-identifier values for the benefit of TypeBox's validators. But I don't think I can accept because updates to TypeBox's compositor would need a symmetric type-level implementation to match the runtime mapping (meaning the `escapeRegexSpecialChars` would need to be expressed the same way in the type system). 

In general, the effect of not having symmetric mapping tends to pop up as downstream inference issues. Any string manipulation (particularly when related to keys, and especially for a shared function), tends to lead to non-trivial fixes needed downstream (or `options` propagated through compositor mappings which is a no go pattern for TB). As a general rule, I have found it better to leave mapping behaviors in an symmetric (albeit incorrect) state (which leaves room for a more systemic fix later)

(fwiw, I think it's questionable that the TB validators are calling into compositor mappings. They shouldn't really be doing that so I think this aspect could use a review in and of itself)

All up, I do think a fix is required here, but a solution needs a more system level review as to how TypeBox is handling key resolution and a review understanding implications of that downstream (validation). There is a new version of TypeBox due out late September (tentative), and I have added your PR repro case for investigation there (there's are a lot of updates being prepared, and a lot more flexibility to change internal mapping behaviors, so will fix things there.

Sorry about that, and the delay. but thanks for reporting the original issue. I will push for a fix in the next minor revision.
Cheers
S

PR #1273, opened 2025-06-20
fix: correctly score nested unions

### LINKED ISSUE #1268: Value.Cast with nested Type.Union produces incorrect result
I have the following schema:

```ts
const result = Value.Cast(
  Type.Union([
    Type.Union([
      Type.Object({
        type: Type.Literal('a'),
        name: Type.String(),
        in: Type.String(),
      }),
      Type.Object({
        type: Type.Literal('b'),
        description: Type.Optional(Type.String()),
        nested: Type.Object({
          a: Type.String(),
          b: Type.Optional(Type.String()),
        }),
      }),
    ]),
    Type.Object({
      $ref: Type.String(),
      description: Type.Optional(Type.String()),
    }),
  ]),
  {
    type: 'b',
    description: 'Hello World',
    nested: {
      b: 'hello',
    },
  },
)

// ❌ Output: { "$ref": "",  "description": "Hello World" }
// ✅ Expected: {"type":"b","description":"Hello World","nested":{"a":"","b":"hello"}}
console.log(result)
```

This behaviour seams to only happen when we have nested union types. When we remove the union we get the correct result

```ts
const result = Value.Cast(
  Type.Union([
    Type.Object({
      type: Type.Literal('b'),
      description: Type.Optional(Type.String()),
      nested: Type.Object({
        a: Type.String(),
        b: Type.Optional(Type.String()),
      }),
    }),
    Type.Object({
      $ref: Type.String(),
      description: Type.Optional(Type.String()),
    }),
  ]),
  {
    type: 'b',
    description: 'Hello World',
    nested: {
      b: 'hello',
    },
  },
)

// ✅ Output: {"type":"b","description":"Hello World","nested":{"a":"","b":"hello"}}
console.log(result)
```

DemonHa:
**✨ Summary**

This PR improves the type scoring system to properly handle nested union types, which were previously ignored by the scoring system

**Problem**

Currently, the scoring algorithm does not account for union types that are nested within other unions or types. This results in inaccurate or misleading scores when comparing complex type structures.

**Solution**

This PR introduces a recursive scoring mechanism that traverses and evaluates all nested union types. Each branch of the union is now scored independently and factored into the final score, improving the overall accuracy and reliability of the scoring system.

closes #1268

### DIFF
--- a/src/value/cast/cast.ts
+++ b/src/value/cast/cast.ts
@@ -76,6 +76,11 @@ function ScoreUnion(schema: TSchema, references: TSchema[], value: any): number 
       const exists = keys.includes(key) ? point : 0
       return acc + (literal + checks + exists)
     }, 0)
+  } else if (schema[Kind] === "Union") {
+    const schemas = schema.anyOf.map((schema: TSchema) =>
+      Deref(schema, references)
+    )
+    return  Math.max(...schemas.map((schema: TSchema) => ScoreUnion(schema, references, value)))
   } else {
     return Check(schema, references, value) ? 1 : 0
   }
@@ -127,7 +132,7 @@ function FromArray(schema: TArray, references: TSchema[], value: any): any {
 function FromConstructor(schema: TConstructor, references: TSchema[], value: any): any {
   if (Check(schema, references, value)) return Create(schema, references)
   const required = new Set(schema.returns.required || [])
-  const result = function () {}
+  const result = function () { }
   for (const [key, property] of Object.entries(schema.returns.properties)) {
     if (!required.has(key) && value.prototype[key] === undefined) continue
     result.prototype[key] = Visit(property as TSchema, references, value.prototype[key])
--- a/test/runtime/value/cast/union.ts
+++ b/test/runtime/value/cast/union.ts
@@ -161,4 +161,190 @@ describe('value/cast/Union', () => {
     Assert.IsEqual(Value.Cast(RA, [A, B], { type: 'A' }), { type: 'A' })
     Assert.IsEqual(Value.Cast(RB, [A, B], { type: 'A' }), { type: 'A' })
   })
+
+  // ------------------------------------------------------------------------
+  // ref: https://github.com/sinclairzx81/typebox/issues/1268
+  // ------------------------------------------------------------------------
+  it('should correctly score nested union types #1', () => {
+    const A =
+      Type.Union([
+        Type.Union([
+          Type.Object({
+            type: Type.Literal('a'),
+            name: Type.String(),
+            in: Type.String(),
+          }),
+          Type.Object({
+            type: Type.Literal('b'),
+            description: Type.Optional(Type.String()),
+            nested: Type.Object({
+              a: Type.String(),
+              b: Type.Optional(Type.String()),
+            }),
+          }),
+        ]),
+        Type.Object({
+          $ref: Type.String(),
+          description: Type.Optional(Type.String()),
+        }),
+      ],
+
+      );
+
+    Assert.IsEqual(Value.Cast(A, {
+      type: 'b',
+      description: 'Hello World',
+      nested: {
+        b: 'hello',
+      },
+    }), {
+      type: 'b',
+      description: 'Hello World',
+      nested: { a: '', b: 'hello' },
+    });
+  });
+
+  it('should correctly score nested union types #2', () => {
+    const A = Type.Union([
+      Type.Union([
+        Type.Object({
+          prop1: Type.String(),
+          prop2: Type.String(),
+          prop3: Type.String(),
+        }),
+        Type.Object({
+          prop1: Type.String(),
+          prop4: Type.String(),
+          prop5: Type.String(),
+        })
+      ]),
+      Type.Union([
+        Type.Object({
+          prop6: Type.String(),
+          prop7: Type.String(),
+          prop8: Type.String(),
+        }),
+        Type.Object({
+          prop1: Type.String(),
+          prop9: Type.String(),
+          prop10: Type.String(),
+        }),
+      ]),
+    ])
+
+    // Picks the first union variant when the score is equal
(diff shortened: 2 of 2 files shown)


sinclairzx81:
@DemonHa Hi, this is a good PR, thank you.

Can you just add a couple more test cases for this? Ideally I want to see tests for structures like.


### Case A

```typescript
type A = TUnion<[
  TUnion<[
    TObject<...>,
    TObject<...>
  ],
  TUnion<[
    TObject<...>,
    TObject<...>
  ]>
]>
```

### Case B

```typescript
type B = TUnion<[
  TObject<...>,
  TObject<...>,
  TUnion<[
    TObject<...>,
    TObject<...>
  ]>
]>
```
Also, just put a heading above these tests

```typescript
// ------------------------------------------------------------------------
// ref: https://github.com/sinclairzx81/typebox/pull/1273
// ------------------------------------------------------------------------
it('should correctly score nested union types 1', () => {
  // test here
})
it('should correctly score nested union types 2', () => {
  // test here
})
it('should correctly score nested union types 3', () => {
  // test here
})
```
This just helps me cross reference this logic back to the original PR. 

---

This PR looks good but given the nature of the update, just want some additional tests around it to assert the expected behavior. Drop in as many tests as you can for as many structures as you can think of. Once their in, can merge this one through.

Thanks again!
S

DemonHa:
@sinclairzx81 Just noticed another issue with the rewarding system

Scoring system is rewarding properties matches 1/propertyCount. This would result on incorrect result (see example below)
```ts
it.only('should correctly reward, () => {
  const A = Type.Union([
    Type.Object({
      prop1: Type.String(),
      prop2: Type.String(),
      prop3: Type.String(),
    }),
    Type.Object({
      prop1: Type.String(),
      prop2: Type.String(),
      prop4: Type.String(),
      prop5: Type.String(),
      prop6: Type.String(),
      prop7: Type.String(),
      prop8: Type.String(),
      prop9: Type.String(),
      prop10: Type.String(),
    }),
  ])

  Assert.IsEqual(Value.Cast(A, {
    prop1: '',
    prop2: '',
    prop7: ''
  }), {
      prop1: '',
      prop2: '',
      prop4: '',
      prop5: '',
      prop6: '',
      prop7: '',
      prop8: '',
      prop9: '',
      prop10: '',
  });
})
```

Is this a design choice?

sinclairzx81:
Sorry, accidently closed this via another PR (re-opened)

DemonHa:
> @sinclairzx81 Just noticed another issue with the rewarding system
> 
> Scoring system is rewarding properties matches 1/propertyCount. This would result on incorrect result (see example below)
> 
> ```ts
> it.only('should correctly reward, () => {
>   const A = Type.Union([
>     Type.Object({
>       prop1: Type.String(),
>       prop2: Type.String(),
>       prop3: Type.String(),
>     }),
>     Type.Object({
>       prop1: Type.String(),
>       prop2: Type.String(),
>       prop4: Type.String(),
>       prop5: Type.String(),
>       prop6: Type.String(),
>       prop7: Type.String(),
>       prop8: Type.String(),
>       prop9: Type.String(),
>       prop10: Type.String(),
>     }),
>   ])
> 
>   Assert.IsEqual(Value.Cast(A, {
>     prop1: '',
>     prop2: '',
>     prop7: ''
>   }), {
>       prop1: '',
>       prop2: '',
>       prop4: '',
>       prop5: '',
>       prop6: '',
>       prop7: '',
>       prop8: '',
>       prop9: '',
>       prop10: '',
>   });
> })
> ```
> 
> Is this a design choice?

Should I create another issue for this?

sinclairzx81:
@DemonHa 

> Scoring system is rewarding properties matches 1/propertyCount. This would result on incorrect result (see example below)

I don't think there is a perfectly ideal way to score schematics against values. The current algorithm scores matching discriminator fields as more likely to contribute to a match, but there is still potential for excessive property counts to counter weight discriminator scores.

The logic for scoring is actually really old (it was one of the first bits of logic written for the `value` submodule). I have been meaning to pull this logic out into a separate module, I wonder if it would be beneficial to pull the scoring logic out from Cast, and into it's own module.

```typescript
import { MatchVariantFuzzy } from '@sinclair/typebox/value'

// where A, B, C and D are schemas / variants
// where index is 0-N on match, or -1 on no match
const selectedIndex = SelectVariantFuzzy([A, B, C, D], value) 
```
If you see potential to improve upon the algorithm, it might be easier to implement external to TypeBox first, then test it, then it could be integrated as a new value submodule. Possibly.

```typescript
./src/value/match/exact.ts // matches union variant exactly (via check)
./src/value/match/fuzzy.ts // matches union variant using fuzzy scoring system
```
Having these separated out from Cast would let this logic be useful elsewhere. 

Thoughts?

DemonHa:
> I don't think there is a perfectly ideal way to score schematics against values. The current algorithm scores matching discriminator fields as more likely to contribute to a match, but there is still potential for excessive property counts to counter weight discriminator scores.

@sinclairzx81 Can we move this discussion to another thread?  I will open another issue on the scoring system

sinclairzx81:
> Can we move this discussion to another thread? I will open another issue on the scoring system

Sure, but let's keep it as a discussion thread rather than an new issue. Given the nature of algorithm, a general discussion would be a better forum to formalize ways to improve upon the current approach.

---

As for this PR, it looks good (and thanks for the additional tests). I'll give the updates a more in depth review tonight and look at merging through either tonight or tomorrow.

Thanks again!

DemonHa:
> Sure, but let's keep it as a discussion thread rather than an new issue. Given the nature of algorithm, a general discussion would be a better forum to formalize ways to improve upon the current approach.

I'll try to improve on the current scoring system and open another PR for that. Thank you!

sinclairzx81:
@DemonHa This update looks good. Will merge :)

sinclairzx81:
@DemonHa Updates published on 0.34.37 

Thanks again for the PR :)
S

PR #1282, opened 2025-06-30
Preserve exact type matches in union conversion

### LINKED ISSUE #1281: Union type conversion issue with Value.Parse()
When using `Value.Parse()` with union types that include both `String` and `Number`, numeric values are unexpectedly converted to strings, even when the original value is already a valid number type.

The conversion behavior seems to prioritise the order of types in the union rather than preserving the original type when it's already valid.

## Expected Behavior

`Value.Parse()` should preserve the original type of the value when it matches one of the union types, rather than converting it to the first matching type in the union.

## Actual Behavior

Numeric values are converted to strings when `Type.String()` appears before `Type.Number()` in the union definition.

## Reproduction Code

```typescript
import { Type, Value } from '@sinclair/typebox';

// Schema with String before Number in union
const schemaStringFirst = Type.Object({
  data: Type.Array(
    Type.Record(
      Type.String(),
      Type.Union([Type.String(), Type.Number(), Type.Null()]),
    ),
  ),
});

// Schema with Number before String in union  
const schemaNumberFirst = Type.Object({
  data: Type.Array(
    Type.Record(
      Type.String(),
      Type.Union([Type.Number(), Type.String(), Type.Null()]),
    ),
  ),
});

// Test data
const testData = {
  data: [
    { key1: "hello", key2: 42, key3: null }
  ]
};

// This converts 42 to "42"
const result1 = Value.Parse(schemaStringFirst, testData);
console.log(result1.data[0].key2); // "42" (string) - UNEXPECTED

// This preserves 42 as number
const result2 = Value.Parse(schemaNumberFirst, testData);  
console.log(result2.data[0].key2); // 42 (number) - EXPECTED
```

## Root Cause

The issue appears to be with the default `Convert` operation in `Value.Parse()`. The conversion logic tries to coerce values to the first compatible type in the union, rather than checking if the value already matches one of the types exactly.

## Workaround

Explicitly specify the operations array without `Convert`:

```typescript
const result = Value.Parse(
  ["Clone", "Clean", "Default", "Assert", "Decode"],
  schema,
  value
);
```

## Environment

- TypeBox version: 0.34.37
- Node.js version: 22.12.0
- TypeScript version: 5.5.4

iamramtin:
## Description

Fixes union type conversion issue where `Value.Parse()` unnecessarily converts values to the first matching type in a union, even when the input already matches one of the union types exactly.

## Problem

When using unions like `Type.Union([Type.String(), Type.Number()])`, numeric values were being converted to strings because the conversion logic would coerce to the first compatible type rather than preserving the original when it's already valid.

```typescript
// Before (broken)
const schema = Type.Union([Type.String(), Type.Number()])
Value.Parse(schema, 42) // Returns "42" (string) instead of 42 (number)

// After (fixed)  
Value.Parse(schema, 42) // Returns 42 (number), which correctly preserves type
```

Fixes #1281

### DIFF
--- a/src/value/convert/convert.ts
+++ b/src/value/convert/convert.ts
@@ -243,6 +243,14 @@ function FromUndefined(schema: TUndefined, references: TSchema[], value: any): u
   return TryConvertUndefined(value)
 }
 function FromUnion(schema: TUnion, references: TSchema[], value: any): unknown {
+  // Check if original value already matches one of the union variants
+  for (const subschema of schema.anyOf) {
+    if (Check(subschema, references, value)) {
+      return value
+    }
+  }
+
+  // Attempt conversion for each variant
   for (const subschema of schema.anyOf) {
     const converted = Visit(subschema, references, Clone(value))
     if (!Check(subschema, references, converted)) continue
--- a/test/runtime/value/convert/union.ts
+++ b/test/runtime/value/convert/union.ts
@@ -40,4 +40,41 @@ describe('value/convert/Union', () => {
     Assert.IsEqual(B, { b: 1, c: 2 })
     Assert.IsEqual(C, { a: 1, b: '2', c: 3 }) // note: matching on first
   })
+  it('Should preserve number type in string-first union when value already matches', () => {
+    const T = Type.Union([Type.String(), Type.Number()])
+    const numValue = 42
+    const strValue = 'hello'
+
+    const A = Value.Convert(T, numValue)
+    const B = Value.Convert(T, strValue)
+
+    Assert.IsEqual(typeof A, 'number')
+    Assert.IsEqual(typeof B, 'string')
+    Assert.IsEqual(A, 42)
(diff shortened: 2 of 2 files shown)


sinclairzx81:
@iamramtin Hi, thanks for the PR 

Hey sorry about the delay getting this PR looked at. I have been extraordinarily busy on other project work of late which has meant I have had to defer PR reviews, sorry about that.

This said, everything looks great here (and this is a nice fix to improve things for order independent unions!). Happy to merge this one in. 

Will publish out later today and let you know once published on this thread.
Thanks again!

sinclairzx81:
@iamramtin Hiya,

Updates published on 0.34.38.

Thanks again for the PR, and again, sorry for the delay.
All the best!
S

iamramtin:
Hey @sinclairzx81, all good, thanks for merging this in. I appreciate the work you're doing here with typebox!

Issue #1292, opened 2025-08-14
Union type scoring incorrectly favors schemas with fewer properties


DemonHa:
When casting a value against a union of object schemas, the scoring logic appears biased toward schemas with fewer properties. This results in incorrect selection when the input matches multiple candidates but one has more properties than the other.

```ts
const schema = Type.Union([
  Type.Object({
    summary: Type.Optional(Type.String()),
    description: Type.Optional(Type.String()),
    parameters: Type.Optional(Type.Array(Type.Any())),
    responses: Type.Optional(Type.Record(Type.String(), Type.Any())),
    requestBody: Type.Optional(Type.Any()),
  }),
  Type.Object({
    $ref: Type.String(),
    summary: Type.Optional(Type.String()),
  }),
])

const result = Value.Cast(schema, {
  summary: 'Get all users',
  parameters: {},
})

// ❌ Output:            { '$ref': '', summary: 'Get all users' }
// ✅ Expected output:   { summary: 'Get all users', parameters: [] }
console.log(result)
```

Expected behavior:
The scoring should prioritize the schema that best matches the provided properties, regardless of how many total properties are defined in the schema. In this case, the first object schema matches more closely and should be selected.

Impact:
This affects scenarios where unions contain both rich object definitions and simpler $ref schemas, leading to incorrect casting.

sinclairzx81:
Closed via https://github.com/sinclairzx81/typebox/pull/1298

PR #1293, opened 2025-08-14
fix: union scoring bias toward schemas with fewer properties

DemonHa:
**Problem**

When casting values against a union of object schemas, the scoring algorithm incorrectly favored schemas with fewer properties. This caused situations where partially matching simple schemas (e.g., $ref objects) were chosen over more accurate matches with richer schemas, leading to incorrect output.

See #1292 

**Solution**

- Updated union scoring to select the schema that best matches the provided properties, rather than favoring schemas with fewer properties.

- Implemented weighted scoring for object property matching:
   - Literal type matches now count as 10× more important in the score.
   - Properties that exist but have the wrong type now count as ¹⁄₁₀th the value of a valid match.

### DIFF
--- a/src/value/cast/cast.ts
+++ b/src/value/cast/cast.ts
@@ -57,23 +57,21 @@ export class ValueCastError extends TypeBoxError {
   }
 }
 // ------------------------------------------------------------------
-// The following will score a schema against a value. For objects,
-// the score is the tally of points awarded for each property of
-// the value. Property points are (1.0 / propertyCount) to prevent
-// large property counts biasing results. Properties that match
-// literal values are maximally awarded as literals are typically
-// used as union discriminator fields.
+// The following logic assigns a score to a schema based on how well it matches a given value.
+// For object types, the score is calculated by evaluating each property of the value against
+// the schema's properties. To avoid bias towards objects with many properties, each property
+// contributes equally to the total score. Properties that exactly match literal values receive
+// the highest possible score, as literals are often used as discriminators in union types.
 // ------------------------------------------------------------------
 function ScoreUnion(schema: TSchema, references: TSchema[], value: any): number {
   if (schema[Kind] === 'Object' && typeof value === 'object' && !IsNull(value)) {
     const object = schema as TObject
     const keys = Object.getOwnPropertyNames(value)
     const entries = Object.entries(object.properties)
-    const [point, max] = [1 / entries.length, entries.length]
     return entries.reduce((acc, [key, schema]) => {
-      const literal = schema[Kind] === 'Literal' && schema.const === value[key] ? max : 0
-      const checks = Check(schema, references, value[key]) ? point : 0
-      const exists = keys.includes(key) ? point : 0
+      const literal = schema[Kind] === 'Literal' && schema.const === value[key] ? 100 : 0
+      const checks = Check(schema, references, value[key]) ? 10 : 0
+      const exists = keys.includes(key) ? 1 : 0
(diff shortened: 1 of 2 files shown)


amritk (approved):
Excellent approach.

@sinclairzx81 if you want perfect matches to be scored higher we can just add something for perfect matches, but I think just ordering the schema you want first in the case of a tie is fine!

sinclairzx81:
@DemonHa Hey, thanks for the PR! I think this one looks good.

### Scoring

The original fuzzy scoring system was based on the idea that two equally sized (entry length) types would differentiate if incrementing the score based on 1/length, but yeah, doesn't work so well with vastly different sized objects. 

If I recall, I had anticipated a potential for lengths to sometimes override literal matches, but I don't think there would be many cases where one would expect a literal property to be de-prioritized due to entry lengths, so I think your update to discard the 1/size scoring in favor of simple fix sized scoring is going to lead to more intuitive select. 

Ill need a couple of days to sit down with this PR and go through a couple of edge cases, but I expect this PR to be merged in over the weekend. It seems ok.

---

### TypeBox Next

Just on the topic of `Value.Cast`. There is actually a new major (or minor) version of TypeBox due for release soon. The `Value.Cast` function is being renamed to `Value.Repair` which is a bit more reflective of the functions original purpose was (i.e. data migration). 

The new version is being prepared for a mid to late September release. So, given the name change (which is breaking), this is a great opportunity to update the logic behind it, so this PR is really good timing, but just a heads up on the naming change.

---

Will report back soon
Thanks Again
S

DemonHa:
@sinclairzx81 

**Additional Note:**

In [this discussions](https://github.com/sinclairzx81/typebox/discussions/1277) you suggested normalizing the values.

However, I believe this step can be omitted. Because the type in question is a union, it is not possible to match both a primitive type and an object type simultaneously. Consequently, comparisons will always occur between values of the same type or compatible subtypes. This allows us to forgo normalizing the output to a fixed range (e.g., 0–1).

In addition, if the value matches the literal type in the object property, we assign it 100 points, which is 10 times the weight of a standard match (10 points). Conversely, if the property exists but is of the wrong type, it is assigned only 1 point—10% of a standard match. These weighting values may be adjusted further if necessary.

Finally, in cases where there is an exact match of the properties, the algorithm will now select the first element in the union type rather than favoring the object with fewer properties. This approach is more intuitive and predictable, allowing developers to prioritize types by ordering them accordingly.

sinclairzx81:
@DemonHa Heya,

Re: normalization

> However, I believe this step can be omitted. Because the type in question is a union, it is not possible to match both a primitive type and an object type simultaneously. Consequently, comparisons will always occur between values of the same type or compatible subtypes. This allows us to forgo normalizing the output to a fixed range (e.g., 0–1).

Yes that's fine. I do quite like normalization if it relates to probability (where each variant union yields a 0-1), but it should be possible to handle normalization external to the actual scoring system (where it's probably better to have more simple integer score tally), so that's fine.

```typescript
// this is fine external to the actual scoring system.
const scores = T.anyOf.map((type) => ScoreVariant(type, value))
const total = scores.reduce((result, score) => result + score, 0)
const normal = scores.map(score => score / total)
```

All good, lets leave out normalization

sinclairzx81:
@DemonHa Hiya!

All things looked good here so have merged + published out on 0.34.40. Excellent work! Also, thanks again for your contribution to this aspect of TB, this update was a good one, it's really appreciated :)

All the best!
S

Issue #1300, opened 2025-08-24
`Value.Clone` does not work with circular structures


DemonHa:
**Problem**
The current Clone function does not handle recursive (circular) structures. When an object references itself, the function runs into infinite recursion and throws a Maximum call stack size exceeded error.

**Example**
```ts
const a = { x: 1 };
a.b = a;

Clone(a); // ❌ Fails with Maximum call stack error
```

sinclairzx81:
@DemonHa Hi, Thanks for your work on this and associated PR's.

## Provision

At this stage, I am looking to hold off on updates to Value.* as there is a new version of TypeBox set for release next month which has refactored the Value submodule entirely. For now, I am keen to lock down functional changes to keep Value functionality in sync between this version and the next. Any significant new functionality will require syncing with the upstream version being prepared for release, so changes like this are best deferred until the new version is out.

## Cyclic

In terms of Cyclic handling, I'm not too convinced TypeBox should handle this. Cyclics have been considered in the past, but given their nature, their introduction would add a significant amount of complexity and performance overhead (reference tracking) that can mostly be avoided using reference pointers encoded in the data itself (i.e. JSON Pointer RFC 6901).

## Clone

I am somewhat open to supporting cyclics in Clone. This would align with `structuredClone` behaviors, and I think the WeakRef tracking would be acceptable in this context.

```typescript
const A: any = { B: null } 

A.B = A

// ....

const X = structuredClone(A)    // This is fine

const Y = JSON.stringify(A)     // TypeError: Converting circular structure to JSON
```

---

## Check and Cast

In terms of Cast and Check, supporting cyclics is probably going to be infeasible. The next version of TypeBox will be moving to a pure Json Schema validation backend. As JSON does not support cyclic referencing, JSON schema has no concept of cyclics. In this regard, updating Check and Cast to support cyclics may prove awkward under a specification regime that has no concept of them.  

In addition, the performance hit for cyclic tracking will almost certainly be too expensive. It would require 1 x WeakRef allocation per visited object where every object has the potential to be cyclic. If TypeBox were to support this, cyclic tracking would need to be optional and configured behind a flag. The cost of supporting them by default would implicate the 99.999% of data structures that are not cyclic, so cyclics should be considered the exception, not the rule.

---

[...]

PR #1301, opened 2025-08-24
feat: support recursive structures

DemonHa:
The current Clone function does not handle recursive (circular) structures. When an object references itself, the function runs into infinite recursion and throws a Maximum call stack size exceeded error.

See issue: #1300

```ts
const a = { x: 1 };
a.b = a;

Clone(a); // ❌ Fails with Maximum call stack error
```

This PR adds support for cloning recursive structures by detecting and handling circular references. Objects that reference themselves (directly or indirectly) can now be cloned safely without causing stack overflows.

### DIFF
--- a/src/value/clone/clone.ts
+++ b/src/value/clone/clone.ts
@@ -35,18 +35,26 @@ import { IsArray, IsDate, IsMap, IsSet, IsObject, IsTypedArray, IsValueType } fr
 // ------------------------------------------------------------------
 // Clonable
 // ------------------------------------------------------------------
-function FromObject(value: FromObject): any {
+function FromObject(value: FromObject, cache: WeakMap<object, unknown>): any {
+  if (cache.has(value)) return cache.get(value)
   const Acc = {} as Record<PropertyKey, unknown>
+  cache.set(value, Acc)
   for (const key of Object.getOwnPropertyNames(value)) {
-    Acc[key] = Clone(value[key])
+    Acc[key] = Clone(value[key], cache)
   }
   for (const key of Object.getOwnPropertySymbols(value)) {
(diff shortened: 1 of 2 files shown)


sinclairzx81:
@DemonHa Hi, just to give an update here. 

So, I spent some time with this during the 1.0 version update, but I am concerned about the performance hit caused by `<ref *>` tracking. As a reference, if we compare the performance of Value.Clone vs structuredClone, there is an order of magnitude difference which is largely a consequence of structuredClone tracking recursive ref, transferrable and other structures. 

```typescript
import Value from 'typebox/value'

const A = { x: 1, y: 2, z: 3 }

console.time()
for(let i = 0; i < 1_000_000; i++) Value.Clone(A)     // 348ms
console.timeEnd()

console.time()
for(let i = 0; i < 1_000_000; i++) structuredClone(A) // 3041ms !!!
console.timeEnd()
```

I should provision that the Clone function in TypeBox is somewhat performance sensitive, so degrading performance by 10x just to support a relatively uncommon data structure is not going to be a good trade off, especially as `<ref *>` cyclics cannot be represented in JSON (the primary data format this library was written for)

I think I am going to close out this PR as a result, sorry about this. I think at this stage, I'm going to put cyclic self referential values in the "out of scope" category, as supporting them for Clone would also imply support for validation, which implies a need to `<ref *>` track there also. This said, TypeBox does support singular and mutual recursive types, but is limited to recursive values that terminate, not non-terminating cyclic values that require ref tracking (which should generally be avoided in data formats anyway)

So, sorry about this. I may consider this functionality in future, at this stage, the cost of taking this one would impact too much. If you need to clone cyclic values, the built in structuredClone function should work.

Will close this one out.
Thanks again
S

Issue #1302, opened 2025-08-25
`Value.Check` breaks on recursive structures


DemonHa:
**Problem**

`Value.Check` fails when validating recursive structures, leading to a maximum call stack exceeded error.

For example:
```ts
const value = {
  id: "1",
  nodes: [],
};

// @ts-expect-error
value.nodes[0] = value;

Value.Check(T, value); // ❌ Callstack error
```

sinclairzx81:
Closing with reference to comment https://github.com/sinclairzx81/typebox/issues/1300#issuecomment-3226770383

PR #1303, opened 2025-08-25
feat: Value.Check support for recursive structures

DemonHa:
The current Check function does not handle recursive (circular) structures. When an object references itself, the function runs into infinite recursion and throws a Maximum call stack size exceeded error.

See issue: https://github.com/sinclairzx81/typebox/issues/1302

```ts
const value = {
  id: '1',
  nodes: [],
}

// @ts-expect-error
value.nodes[0] = value

Value.Check(T, value) // ❌ Fails with Maximum call stack error
```

This PR adds support for checking recursive structures by detecting and handling circular references. Objects that reference themselves (directly or indirectly) can now be checked safely without causing stack overflows.

### DIFF
--- a/src/value/check/check.ts
+++ b/src/value/check/check.ts
@@ -109,7 +109,7 @@ function FromAny(schema: TAny, references: TSchema[], value: any): boolean {
 function FromArgument(schema: TArgument, references: TSchema[], value: any): boolean {
   return true
 }
-function FromArray(schema: TArray, references: TSchema[], value: any): boolean {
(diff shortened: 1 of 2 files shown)


sinclairzx81:
Closing with reference to comment https://github.com/sinclairzx81/typebox/issues/1300#issuecomment-3226770383

Note, the tests implemented for this PR are valuable. They may be worth keeping to serve as a spec for introducing this functionality in later versions.

PR #1309, opened 2025-08-28
feat: support for recursive structures

DemonHa:
The current `Clone`, `Check` and `Cast` functions does not handle recursive (circular) structures. When an object references itself, the function runs into infinite recursion and throws a Maximum call stack size exceeded error.

```
const a = { x: 1 };
a.b = a;

Clone(a); // ❌ Fails with Maximum call stack error
```

This PR adds support for recursive structures by detecting and handling circular references. Objects that reference themselves (directly or indirectly) can now be processed safely without causing stack overflows.

### DIFF
--- a/.github/workflows/publish-npm.yml
+++ b/.github/workflows/publish-npm.yml
@@ -1,0 +1,24 @@
+name: Publish Package to npmjs
+on:
+  push:
+    tags:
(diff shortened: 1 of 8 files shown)


DemonHa:
It was an accident

PR #1319, opened 2025-09-10
Fix decoding of optional object properties

### LINKED ISSUE #1318: Decoding object with optional properties fails
I'm hitting the "Unreachable" case here: https://github.com/sinclairzx81/typebox/blob/a1417da62f64a69c67f00d65e0909138369d0fe6/src/value/codec/from-object.ts#L45-L51

When attempting to use `Value.Decode(...)` on an object type with optional properties (where one of those properties is missing).

For comparison, the pre-1.0 decoder had special casing for optional properties:
https://github.com/sinclairzx81/typebox/blob/e0ec98c9dbbf6f6b8777f121a2da38f2d09cf2ed/src/value/transform/decode.ts#L134-L145

Diggsey:
Fixes #1318


sinclairzx81:
Closing as per comment https://github.com/sinclairzx81/typebox/issues/1318#issuecomment-3275836351

PR #1346, opened 2025-09-20
Declare the cause field on AssertError instances

webstrand:
This makes it cleaner to extract the errors field from the exception.

### DIFF
--- a/src/value/assert/assert.ts
+++ b/src/value/assert/assert.ts
@@ -35,6 +35,7 @@ import { Errors } from '../errors/index.ts'
 // AssertError
 // ------------------------------------------------------------------
 export class AssertError extends Error {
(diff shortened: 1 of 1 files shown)


sinclairzx81:
@webstrand Hey, thanks :)

I think this is a good idea. I wasn't sure whether `cause` should be opaque given it's not specified in default `.lib.ts` files, but is used to present error values in Node/Deno. The alternative would be to duplicate the assertion parameters as fields, but wasn't keen on that either. 

I like this this approach so will merge through. I will apply a couple of additional updates and publish out on 1.0.15.

Thanks again!
S

sinclairzx81:
@webstrand Heya,

Update published on 1.0.15. Thanks again!

All the best
S

PR #1357, opened 2025-09-24
Add Type.Convert to control Value.Convert behavior

webstrand:
Add a `Type.Convert` extension that allows modifying or overriding `Value.Convert` behavior. Similar to `Decode`, except this transforms values _before_ validation.

## Open questions
1. Is `Type.Convert` an appropriate name for this extension.
2. Is the object literal `ConvertCallbackResult` appropriate, or should constructor functions be provided?
3. Is splitting `FromType` into `FromType` and `FromTypeWithoutConvert` appropriate or should a boolean flag be used instead?
4. Implemented behavior each `Type.Convert(Inner)` wrapper prepends its callbacks to the list to be executed, is this appropriate?

### Extended coercion rules
I am using TypeBox to parse data loaded from a directory tree. It would be wonderful if I could extend `Value.Convert` to understand how to coerce `Buffer` objects into strings and numbers.
```ts
const Buffer = {
  asString(options: Type.TStringOptions) { 
    return Type.Convert(
      Type.String(options), 
      (value) => ({ action: "continue", value: Buffer.isBuffer(value) ? value.toString() : value })
    ) 
  },
  asNumber(options: Type.TNumberOptions) { 
    return Type.Convert(
      Type.Number(options), 
      (value) => ({ action: "continue", value: Buffer.isBuffer(value) ? value.toString() : value })
    ) 
  }
}

envdir("env", Type.Object({
    cert: Buffer.asString({ pattern: "BEGIN" }),
    port: Buffer.asNumber({ minimum: 0, multipleOf: 1, maximum: (1 << 16) - 1 }),
}))
```

### Suppress undesirable coercion
```ts
const A = Type.Object({ example: Type.String() })
const B = Type.Object({ example: Type.Number({ minimum: 5 }) })
const U = Type.Union([A, B])

// Coercion rules result in
Type.Parse(U, { example: 0 })
// succeeding with `{ example: "0" }`, regardless that it would fail to match `B`

const StrictU = Type.Convert(U, (value) => ({ action: "final", value }));
Type.Parse(StrictU, { example: 0 })
// exception
```

### DIFF
--- a/src/type/types/_convert.ts
+++ b/src/type/types/_convert.ts
@@ -1,0 +1,77 @@
+/*--------------------------------------------------------------------------
+
+TypeBox
+
+The MIT License (MIT)
+
+Copyright (c) 2017-2025 Haydn Paterson 
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+
+---------------------------------------------------------------------------*/
+
+// deno-fmt-ignore-file
+
+import { Memory } from '../../system/memory/index.ts'
+import { Guard } from '../../guard/index.ts'
+import { type TSchema, IsSchema } from './schema.ts'
+import { TProperties } from './properties.ts'
+
+// ------------------------------------------------------------------
+// ConvertAdd
+// ------------------------------------------------------------------
+/** Applies a Convert override to the given type. */
+export type TConvertAdd<Type extends TSchema = TSchema> = (
+  '~convert' extends keyof Type ? Type : TConvert<Type>
+)
+/** Applies a Convert override to the given type. */
+export function ConvertAdd<Type extends TSchema>(type: Type, conversion: TConvertCallback<TConvertAdd<Type>>): TConvertAdd<Type> {
+  const conversions = IsConvert(type) ? [conversion, ...type['~convert']] : [conversion]
+  return Memory.Update(type, {  }, { '~convert': conversions }) as never
+}
+// ------------------------------------------------------------------
+// Type
+// ------------------------------------------------------------------
+/** Represents a type with embedded Convert override. */
+export type TConvert<Type extends TSchema = TSchema> = (
+  Type & { '~convert': TConvertCallback<Type>[] }
+)
+// ------------------------------------------------------------------
+// Factory
+// ------------------------------------------------------------------
+export type TConvertCallback<Type extends TSchema> = (value: unknown, type: Type, context: TProperties) => TConvertCallbackResult
+// TODO: Should we use classes here? constructor functions? `(value) => Continue(Number(value) * 100)` or `Final(String(value))`
+export type TConvertCallbackResult = {
+	action: "continue" | "final"
+	value: unknown
+}
+
+/** Applies a Convert override to a type. */
+export function Convert<Type extends TSchema>(type: Type, callback: TConvertCallback<TConvertAdd<Type>>): TConvertAdd<Type> {
+  return ConvertAdd(type, callback) as never
+}
+// ------------------------------------------------------------------
+// Guard
+// ------------------------------------------------------------------
+/** Returns true if the given value is a TConvert. */
+export function IsConvert(value: unknown): value is TConvert<TSchema> {
+  return IsSchema(value) && Guard.HasPropertyKey(value, '~convert') && Guard.IsArray(value["~convert"])
+}
+
+
--- a/src/type/types/index.ts
+++ b/src/type/types/index.ts
@@ -30,6 +30,7 @@ THE SOFTWARE.
 // Extensions
 // ------------------------------------------------------------------
 export * from './_codec.ts'
+export * from './_convert.ts'
 export * from './_optional.ts'
 export * from './_readonly.ts'
 export * from './_refine.ts'
--- a/src/typebox.ts
+++ b/src/typebox.ts
@@ -73,6 +73,7 @@ export { type TUppercase, type TUppercaseDeferred, Uppercase } from './type/acti
 // Extension
 // ------------------------------------------------------------------
 export { Codec, Decode, DecodeBuilder, Encode, EncodeBuilder, IsCodec, type TCodec } from './type/types/_codec.ts'
+export { Convert, IsConvert, type TConvert, type TConvertCallback, type TConvertCallbackResult } from './type/types/_convert.ts'
 export { IsOptional, Optional, type TOptional } from './type/types/_optional.ts'
 export { IsReadonly, Readonly, type TReadonly } from './type/types/_readonly.ts'
 export { IsRefine, Refine, type TRefine, type TRefineCallback, type TRefinement } from './type/types/_refine.ts'
--- a/src/value/convert/from-convert.ts
+++ b/src/value/convert/from-convert.ts
@@ -1,0 +1,47 @@
+/*--------------------------------------------------------------------------
+
+TypeBox
+
+The MIT License (MIT)
+
+Copyright (c) 2017-2025 Haydn Paterson 
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+
+---------------------------------------------------------------------------*/
+
+// deno-fmt-ignore-file
+
+import type { TConvert, TConvertCallbackResult, TProperties } from '../../type/index.ts'
+import { FromTypeWithoutConvert } from './from-type.ts'
+
+export function FromConvert(context: TProperties, type: TConvert, value: unknown): unknown {
+  let action: TConvertCallbackResult["action"] = "continue"
+  for(const conversion of type["~convert"]) {
+    ({action, value} = conversion(value, type, context))
+    if(action === "final") {
+      return value;
+    }
+    else if(action !== "continue") {
+      throw new Error(`Unknown Convert-Callback action type ${action}`)
+    }
+  }
+
+  return FromTypeWithoutConvert(context, type, value);
+}
--- a/src/value/convert/from-type.ts
+++ b/src/value/convert/from-type.ts
@@ -30,6 +30,7 @@ THE SOFTWARE.
 
 import * as T from '../../type/index.ts'
 
+import { FromConvert } from './from-convert.ts'
 import { FromArray } from './from-array.ts'
 import { FromBase } from './from-base.ts'
 import { FromBigInt } from './from-bigint.ts'
@@ -52,6 +53,13 @@ import { FromUnion } from './from-union.ts'
 import { FromVoid } from './from-void.ts'
 
 export function FromType(context: T.TProperties, type: T.TSchema, value: unknown): unknown {
+  return (
+    T.IsConvert(type) ? FromConvert(context, type, value) : 
+    FromTypeWithoutConvert(context, type, value)
+  )
+}
+
+export function FromTypeWithoutConvert(context: T.TProperties, type: T.TSchema, value: unknown): unknown {
   return (
     T.IsArray(type) ? FromArray(context, type, value) :
     T.IsBase(type) ? FromBase(context, type, value) :
--- a/test/typebox/runtime/value/convert/convert.ts
+++ b/test/typebox/runtime/value/convert/convert.ts
@@ -1,0 +1,24 @@
+import { Value } from 'typebox/value'
+import { Type } from 'typebox'
+import { Assert } from 'test'
+
+const Test = Assert.Context('Value.Convert.Convert')
+
+Test('Should Convert 1', () => {
+	const T = Type.Convert(Type.String(), (value) => ({ action: "final", value }))
+	const R = Value.Convert(T, 3.14)
+	Assert.IsEqual(R, 3.14);
+})
+
+Test('Should Convert 2', () => {
+	const T = Type.Convert(Type.String(), (value) => ({ action: "continue", value }))
+	const R = Value.Convert(T, 3.14)
+	Assert.IsEqual(R, "3.14");
+})
+
+Test('Should Convert 3', () => {
+	const T = Type.Convert(Type.String(), function addOne(value) { return  ({ action: "continue", value: Number(value) + 1 }) })
+	const U = Type.Convert(T, function multiplyTen(value) { return  ({ action: "continue", value: Number(value) * 10 }) })
+	const R = Value.Convert(U, 10)
+	Assert.IsEqual(R, "101")
+})
--- a/test/typebox/static/type/convert.ts
+++ b/test/typebox/static/type/convert.ts
@@ -1,0 +1,25 @@
+import { type Static, Type } from 'typebox'
+import { Assert } from 'test'
+
+{
+	const T = Type.Convert(Type.String(), (value) => ({ action: 'continue', value: null }))
+	type T = Static<typeof T>
+
+	Assert.IsExtendsMutual<T, string>(true)
+	Assert.IsExtendsMutual<T, null>(false)
+}
+
+{
+	const T = Type.Convert(Type.String(), (value) => ({ action: 'final', value: null }))
+	type T = Static<typeof T>
+
+	Assert.IsExtendsMutual<T, string>(true)
+	Assert.IsExtendsMutual<T, null>(false)
+}
+
+{
+	const T = Type.Convert(Type.String(), (value, type) => {
+		Assert.IsExtendsMutual<typeof type, typeof T>(true)
+		return { action: 'continue', value: value }
+	})
+}


marcomuser:
I was just looking through the whole documentation to find a preprocessor like Type.Convert and now I see this PR. Would be very useful to be able to transform before validation.

sinclairzx81:
@webstrand Hi, 

Hey, thanks for submitting, and sorry for the delay (busy week!)

---

So, I had a good look through the PR and everything looks great from an implementation stand point. However as mentioned via other channels, I do have a couple of reservations around including additional control keywords (such as `~convert`) into the schema graph as it implies indirect coupling to `Value.*`, but where the long term goal for TB is to try and find ways to remove indirect coupling between `Type.*` from `Value.*`.

This said, I am very much in support of the functionality you're suggesting here, but think additional design work may be required, as any functionality that involves type traversal / interception / control flow for Convert should have equal relevance to Clean, Default, Encode, Decode, etc also.

Have drafted something for consideration below, but will answer your questions first.

---

## Open questions - Responses

> Is Type.Convert an appropriate name for this extension.

Yes, I think Type.Convert would correct naming.

> Is the object literal ConvertCallbackResult appropriate, or should constructor functions be provided?

Constructor functions are required if a caller is exposed to `TConvertCallbackResult` in someway. In this case, we would need `ContinueWith` and `FinalWith` functions to return one of the two possible variant structures. Guards would also need be written internal to the library that check for either variant.

> Is splitting FromType into FromType and FromTypeWithoutConvert appropriate or should a boolean flag be used instead?

I would maybe lean more towards keeping `FromType` as is, but adding a leading `if(~convert)` check to terminate early.

```typescript
export function FromType(context: T.TProperties, type: T.TSchema, value: unknown): unknown {
  // intercept + terminate early
  if(T.IsConvert(type)) return FromConvert(context, type, value) 
  // business as usual
  return ( /* ternary * / )
}
```

> Implemented behavior each Type.Convert(Inner) wrapper prepends its callbacks to the list to be executed, is this appropriate?

Yes, I think so. The Refine and Codec augmentations stack functions in a similar way. I think it makes sense here too.

---

## Reservations

So while I am very much in support of having this functionality, I am still somewhat hesitant to introduce additional non-standard control keywords into schematics as this makes it harder to decouple `Type.*` from `Value.*` (which is a longer term goal for the project). The `~convert` keyword implies some indirect coupling to "something" else in TypeBox, and where the keyword has no meaning to other validators (like Ajv) which introduces a schema portability concern.

> Note: The inclusion of Refine and Codec are somewhat special cases as there is no other way to implement these under current TB architecture. However the longer term goal is to move functions that apply these keywords  to `Value.*`, with the the eventual goal of having `Type.*` ONLY generate schematics with no runtime / function attachments.

So, while I think the functionality you're suggesting for Convert is important (and should be expressed in the library in some capacity), I think the inclusion of this functionality should be handled in a more "abstract" sense and have some applicability to the other `Value.*` functions, specifically, Clean, Default, Create, Repair, Decode, Encode.

## Consideration: Value.Process(...)

Currently, TypeBox implements fixed functions to process actions (Convert, Repair, etc). These functions all share common characteristics in which they:

1) Traverse a type / schematic
2) Inspect a type (keywords + custom keywords)
3) Inspect a value against a current visited type.
4) Apply some action.

These 4 abstract "things" are common to each `Value.*` function and deserve more attention. I think by focusing on these aspects specifically, a solution could be devised to support custom conversions (as per Convert), or any other number of custom operation one might want to apply with a schematic, 

In terms of a approach, the mechanisms the library would be prepared to adopt must support:

1) The ability for users to create their own `~custom` augmentations (including `~convert`)
2) The ability for users to create `Value.*` processors that detect `~custom` augmentations and perform an action.

Based on your Convert implementation, I think you are on the right track with `final` and `continue` return structures as a way to control processing execution. I have drafted extremely provisional `Value.Process` function which incorporates the `final` and `continue` from your Convert function, but tries to generalize things in such a way that Value.Process "could" underpin other `Value.*` functions also.

[Reference Link Here](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAcQK4EMoBM4DMoRHAchgE8wBTAIwgA8B6Ac1QwIChRJY4AVUs7XfEV5VqrFrVpwAtDNlz5CxUuUrV4yQGEUAG20AhFAGMA1uumqLlq9ZYly3DRAB2MYE6RkA6sBgALOAC8cADecHZkAFyEhs6u7mQEADRwAG46HlFITsZOEADuTnAAvra83ACiTujefoEhYbxRBGRVSanpkXBZOfmFJeHcAEpkAM5I2vBBXI4ubh41-gA+FVULLCxYWYauznAzcfM+vgAUadoZXdm5BQCUUdOxc15HISwAkFBkMEhQhaHhTRis3ibTOHmKLBKmyc22Au0q1SOpw6mSuvTuK0RtWC70+31+9QBhBa6FBHQhJQ2Wx2hQAkiN9k8FsMxhNkedOt1rk4MWC+MARpiFq8Pl8foVkGh0AA6ekAeQoACsyNt2R4bu83gAyLWIJgygASKBGAAVcORYMQANJkYhqsjJITkAgat7a3WSjDSo2m81kS02u18x18l2anV6qWykblACOqG09ul4UdJJdcEhVJhNLg9IRzNG4xg9tRPVuUT5cAFQpeONF+Il+ujCuVqr5rvdka9PrNEAtJED9sd4TDbojnsNxt7-ettqHhFDHfHTbz8Z0SZTxNaNwz-TK0x02goRmMdWOQJgZGoMHuvGlXGn-tco2SRJ45HvAGVDL4yCAUMkfKltyO4BAAfEMhYTOsEjmNY8EIfBZgAGICAAglAUAoMQZiIXh+HKFmsK7KheAYVhxC0peIAjOesRXje3B3g+fqwMAL4NOQt4flw36-v+gEopcZZOMkhiHseJj3FouiScYGJcr0Ip4uKXYyvS5HYfaO4APztBy0r-mA9qBBBpEgO+ZB0S4DGvrwgkcmJEknjcO4Vh0mbQsRhTmZpdoXgx3FkPej5sRxb7MX5DkXIpBRObJJ7Sc5Jg7nWQIjPAnyspMcDiQlJjWZe152eQ0VkK6wBYMcearEcLJFscWVFq5cAqQSTUTNKlZwLBl5QCAbgoDmP4qqY9aqb5mHYVRf60QFxWccFPizckHUwF1HTxUeLmZrBBH7QdKECC2KowLhB0XYhRE5uZoXPnN9HXkFIWsfdyRgK97EjM9LF9k+X1lcBvRbXJSX5fJIqVccACEE7Nkqp0AHIQDAfnaTubWFHyuJigScOBrRH1-WFIw3NKnzoEghhWY1UEwMkxi2qBEF1uNeNNj2r2zkGm1wIzxAdm8emhNKotrckADa-MALpROZlmFbZcBE-2X1S7a0tlerxCa7lyUQyUbpvFEa3vEUyTBEUGpQtScI+cdCOqvNjGWS9xNvYtP0ndsgPCdyIOJQ4+upe86WZXTdR5dtBXOyVDr6eq7xQzVWK+PVbJrS1mOtXTG0cj1kh9QNThDXbuW-iYOMNnAt2faMisLeE0oq-9HFrXnHgBylu2SJdfc2LB8u8Od-ej4RXk3QICvOz9d0A57TE8Xxf4AQnnJonFevg2D0fyUDBTKbjhTHJqrsaVNdrhLpNfoRfDf04tZVdxDESn3e8qO8WV9wHp5ne8WsdH5r2fm5DMbpeoQHQBAKIKB0DoB8HbHQi1BS-k+JqCQ6DJCi2lJgzB2BDx+AgEgeg-gpBwFyE4KQMAsIpH9CMFAFBtBkE1FHOS9845lTJtjN41sR5j34XIMwAA1Do0pezUxGLRbBGo9oCLkTIFgV4ODwAnmXcRowHo2SeovYKv1VbhUaDor8I0BJr33qJLeu8d5yRDmze2eBp6PQfpuYMlibGZj4fIseZhhhYH9C0amnivH9xgpIYYRk4CIyQCACg-owgQDgJ+ahbh6AsDDnAdOOV1GSOOJbZIrt-65PeNQZ6USYn+mODcRI7xiClOibEqAlTqlvAAF51PKY0qpkIqn1BKXAAAjMkWpcAABMyQ2lwAAMzFGSOwoB7ZTKH2rifN0Z8RhlIaccb+y4oz0g2RUhZmohZ7EePEBYG4IBJKgCkypO4jlREZGcpEhyeHdPWOlCATDpTaAgPQXJGSI5WyAA)

```typescript
import Guard from 'typebox/guard'
import Type from 'typebox'

// CallBack
type TContinueWith = { type: 'continue', value: unknown }
type TEndWith = { type: 'end', value: unknown }
type TResult = TContinueWith | TEndWith

function ContinueWith(value: unknown): TContinueWith {
	return { type: 'continue', value }
}
function EndWith(value: unknown): TEndWith {
	return { type: 'end', value }
}

function IsContinueWithResult(value: unknown): value is TEndWith {
	return Guard.IsObject(value)
		&& Guard.HasPropertyKey(value, 'type')
		&& Guard.HasPropertyKey(value, 'value')
		&& Guard.IsEqual(value.type, 'end') 
}
function IsEndWithResult(value: unknown): value is TEndWith {
	return Guard.IsObject(value)
		&& Guard.HasPropertyKey(value, 'type')
		&& Guard.HasPropertyKey(value, 'value')
		&& Guard.IsEqual(value.type, 'end') 
}
type TCallback = (context: Type.TProperties, type: Type.TSchema, value: unknown) => TResult

// FromArray
function FromArrayItems(context: Type.TProperties, type: Type.TSchema, value: unknown, callback: TCallback): unknown {
	return Guard.IsArray(value) ? value.map(value => FromType(context, type, value, callback)) : value
}
function FromArray(context: Type.TProperties, type: Type.TArray, value: unknown, callback: TCallback) {
	const result = callback(context, type, value)
	if(IsEndWithResult(result)) return result.value  // termination check
	return FromArrayItems(context, type.items, result.value, callback)
}
// FromObject
function FromProperties(context: Type.TProperties, properties: Type.TProperties, value: unknown, callback: TCallback) {
	if(!Guard.IsObjectNotArray(value)) return value
	return Guard.Keys(properties).reduce((result, key) => {
		return Guard.HasPropertyKey(value, key)
			? { ...result, [key]: FromType(context, properties[key], value[key], callback) }
			: result
	}, {})
}
function FromObject(context: Type.TProperties, type: Type.TObject, value: unknown, callback: TCallback) {
	const result = callback(context, type, value)
	if(IsEndWithResult(result)) return result.value // termination check
	return FromProperties(context, type.properties, result.value, callback)
}
// FromType
function FromType(context: Type.TProperties, type: Type.TSchema, value: unknown, callback: TCallback): unknown {
	return (
		Type.IsArray(type) ? FromArray(context, type, value, callback) :
		Type.IsObject(type) ? FromObject(context, type, value, callback) : 
		// todo: additional types here
		//
		// ...
		//
		// fallthough - non-traversable
		callback(context, type, value).value
	)
}
// Value.Process(...)
export function Process(context: Type.TProperties, type: Type.TSchema, value: unknown, callback: TCallback) {
	return FromType(context, type, value, callback)
}

// Reference

// Remap Number to String
const Result = Process({}, Type.Object({
	x: Type.Number(),
	y: Type.Number(),
	z: Type.Number(),
}), { x: 1, y: 2, z: 3 }, (context, type, value) => {
	return (
		Type.IsNumber(type) && Guard.IsNumber(value) 
			? ContinueWith(value.toString()) 
			: ContinueWith(value) 
	)
})

console.log({ Result })
```

This is a very high level example of what TB would be looking for where each sub schema / type is visited and sent to a user defined callback to apply some mapping to a value. The `EndWith(value)` and `ContinueWith(value)` are inline with your implementation for Convert.

### Important Work!

Solving value traversal and processing in this way is actually really important because if the following can be solved in an abstract sense......

1) Traverse a type / schematic
2) Inspect a type (keywords + custom keywords)
3) Inspect a value against a current visited type.
4) Apply some action.

... It would mean that TypeBox can potentially look at JIT compiling `Value.*` functions, offering high performance Convert, Default, Clean, inline with Check.

---

Hope this helps bring some insight into the thinking going on for Value.*. While I think your implementation is fine, in order to adopt, it would need to be developed more inline with Value.Process, this in support of the following high level architecture.

design

Happy to discuss more!
S

webstrand:
No worries about the delay,

---

I found another union issue while writing this up. I hadn't noticed this one because `Value.Parse` conditionally skips the pipeline and just uses `Check` if it can:

```ts
const A = Type.Object({ a: Type.String() })
const B = Type.Object({ a: Type.String(), b: Type.String() })
const U = Type.Union([A, B])
console.log(Value.Decode(U, { a: "x", b: "x" })) 
// { a: "x" }
```

While overriding `~convert` behavior for `Value.Convert` made sense, it makes much less sense to me to be overriding the behavior of `Value.Clean` through schema keywords. I suspect that `Value.Default` likely has a similar issue.

---

So if I understand correctly, you're suggesting that instead of each `Value.Clean`, `Value.Default`, `Value.Convert`, and `Value.Decode` implementing their own custom traversal of the schema, that they be implemented in terms of a generic visitor, like this?

```ts
Value.Convert = (context, schema, value) => Value.Process(context, schema, value, Visitor.Convert)
```

The TypeBox library would provide premade visitors for the various operations currently in `Value.*`, and expose them on the value namespace as top-level operations?

My concern is: if TypeBox provides a builtin `Value.Parse` which composes these premade operations, users will naturally reach for `Value.Parse` to consume schemas. If I write a schema that needs custom conversion behavior (like the second motivating example where unions need strict matching without coercion), and the schema gets consumed by the builtin `Value.Parse`, it may silently produce incorrect results due to unintended conversions.

This could fragment the ecosystem where schemas with custom keywords only work correctly with custom processing pipelines, breaking interoperability with users that expect to use the standard `Value.Parse`. An alternate solution like adding a keyword that causes TypeBox builtins to reject the schema, might be better? That way users could instead be directed to use the correct custom `Parse` implementation.

---

Perhaps my use-case for the TypeBox library is out of the ordinary. While I enjoy the fact that it's otherwise pure json-schema I don't really intend on interoperating with other libraries or interpreters. I expect users of my library to use `TypeBox` to validate values against the schemas, and if they decide to feed them into another library, like `ajv`, then they're responsible for handling any super-json-schema operations that need to occur for the schema to validate correctly.

sinclairzx81:
@webstrand Hiya,

> This could fragment the ecosystem where schemas with custom keywords only work correctly with custom processing pipelines, breaking interoperability with users that expect to use the standard Value.Parse. An alternate solution like adding a keyword that causes TypeBox builtins to reject the schema, might be better? That way users could instead be directed to use the correct custom Parse implementation.

Actually, there is one other feature in V1 that might allow you to have a portable type (that you can export for users to use), that can intercept processing on Value.Convert (and other Value.*) functions. While it it wasn't specifically written for this purpose, it might actually offer the functionality you need.

## Type.Base

The new Type.Base class/type that went in for V1 has overridable functions that the Value module will call into. This class is TypeBox's answer to [Standard Schema](https://github.com/standard-schema/standard-schema), but interestingly, it has overrides that would technically allow you to intercept a value mid flight, and either continue or return based on a condition.

Here is a quick starter using Base.

```typescript
import Type from 'typebox'
import Value from 'typebox/value'

// Convert
type TConvertCallback = (value: unknown) => unknown
export class TConvert<Type extends Type.TSchema> extends Type.Base<Type.Static<Type>> {
  constructor(private readonly type: Type, private readonly callback: TConvertCallback) {
    super()
  }
  override Check(value: unknown): value is Type.Static<Type> {
    return Value.Check({}, this.type, value)
  }
  override Convert(value: unknown): unknown {
    // intercept, continue based on condition
    // return (continue) ? Value.Convert(this.type, value) : value
    
    return this.callback(value) // or just use callback
  }
}
export function Convert<Type extends Type.TSchema>(type: Type, callback: TConvertCallback): TConvert<Type> {
  return new TConvert(type, callback)
}

// Type
const T = Type.Object({
  x: Convert(Type.Number(), (value) => 1), // embedded control logic
  y: Convert(Type.Number(), (value) => 2),
  z: Convert(Type.Number(), (value) => 3),
})

const R = Value.Parse(T, { x: null, y: null, z: null }) // { x: 1, y: 2, z: 3 }

console.log(R)
```

I had written Base as a way to express non-json encodable types (replacing Date and Uint8Array), but it dawned on me this afternoon (and after spending a few hours trying to integrate with some Standard Schema infrastructure) that Base could also be used to encode control logic in a schema graph in fairly arbitrary ways. 

It's not been tested for this usecase, but might be worth checking out.

sinclairzx81:
@webstrand Hiya, just a quick update on this

So, was experimenting this afternoon using Type.Base as a means to control / terminate Value.* processing.

I came up with the following (which I think should work)

```typescript
// Types + Control

const A = Type.Object({ a: Control.NoConvert(Type.String()) })
const B = Type.Object({ a: Control.NoConvert(Type.String()) , b: Type.String() })
const U = Control.NoClean(Type.Union([A, B]))

// Test

 type U = Static<typeof U>                       // Control Structures Erased on Infer
                                                 //
                                                 // type U = {
                                                 //   a: string;
                                                 // } | {
                                                 //   a: string;
                                                 //   b: string;
                                                 // }

console.log(Value.Decode(U, { a: 'x', b: "x" })) // { a: "x", b: "x" }

console.log(Value.Decode(U, { a: 100, b: "x" })) // Error: [
                                                 //   {
                                                 //     keyword: "~standard",
                                                 //     schemaPath: "#",
                                                 //     instancePath: "",
                                                 //     params: {
                                                 //       vendor: "typebox",
                                                 //       issues: [
                                                 //         { message: "must be string", ... },
                                                 //         { message: "must be string", ... },
                                                 //       ]
                                                 //     },
                                                 //     message: "must match ..."
                                                 //   }
                                                 // ]
```

This is the code for Control.* is below (it's quite tidy)

[TypeScript Link Here](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAFQJ5gKYBo4G84xauAZRgEMZgBjOAXzgDMoIQ4ByPNAIwgA8WAoUJFhwAaiQA2AVwIMmrdqi7cA9ADcJ0-nwoQAdgGd4AQTgBeRPgB0AeQ4ArVBRgAKHCQBccAMJ6YjcZYAchA+uqqosM7IaJbEUMC6AObOAJQpNCnaeoZwAEJmFjG2Dk6ucB7evv5BIXrhkdGosX4JyWlwmByejc3xSakZWQbwAKoFoX4QAcFe4qgkulFWI7rAes4A2kaYuQC6aXx8ysqIqIaHuPhwY+bEZJQAPAoQdNcAfHCfX98-v5-HlV0k3ERD8kickigZzgAFEoCR9KgACZwPRwACSujoET4fzx+IJhO+x1xRLJ5PxAIU1wKWFJFIZDIBnwqhj6iQA3PTGTzCQDaAAfbDc3mi37M8qeNmtLliuXik6fLpwaVJWXyjVwfmHHQGKZNcQQZJiKRNAAijggSNQzhGmDcnhYvE6ngARNxXRl0gCHXB3a6XX6PTQddl9ZZDcaNObLdbbfbJXAAIwABhTgf9Xq1JzhjCgng2Is1vIldOL8olnwA1qgkAB3aBIt0AP0MCyRJCgSIDRfLTMVX30FAAFqgQCQAApkYdugDEPb7YsrcASbd0FFQU5gM79C8XJYHnzAnZIIH0njL+55y8+4V0SOgboUSj3V-7f2A+n00nPcELb8ZG8vhwEAzn0EhElQN0QEkHIOAIVVEgDOBLFQmh0F7ADKUPH4QLAiCoL9GC4IQlokmQ1DLHQzCsL+IDdho2ifmXagMKYsll1Ar8COg2D4HHGARxQ1DXUY9j-gHagxPEgEGL4VBuCEeBdFPM5jw3QFgWFCS4CnL8EGHRhJESYdSQUpS4AocQEX0RAJn8B5GjgBSYFQe9bJ6BBCBHMcSA+Fy3KRDyrFyBFUAeO5yAoRz8DeD5L0+XU2XBGBoGcMB4nUVy4ChEgH10cQkEuNBunwdIcG-NAoAGKTvjASQOHEShUXqeJrW8UcKCrZx1FNTxJF0KtdAgOtdBSTxeukFdgpiSLHkaeKiyhGBIV0URo0sLxOu67dP0sBRMEm1BMm+WqvnqxrmogVrgHa2Z5kWI7+sG4bRvGuABqGka1oSr5ltW9bTU2uYFmcXb9H2-BDujE6vjOo8GqaqhroiNqCFCeoXCej6Xu+97Pten6ltQFaoDWk1pE2uoIhccHIbQaHTVhz54bgC6kZa1HbvR3LXNSUrZtIKKYrQRbfn+snAcprxeZtOmFGZkM6sRq6bvai06BISRxCx6Nnq+t79cJ7Sfgl8mNo1rWdbB4c9oOuAjsV1n2dVrn2tzaB9B6vWcYNsbPAgexHBgDZdhN74zalpoPagL35ahh2Ye5M6zoBABaDPM6z7Oc9zvP84LwvSQBGYQd0YuTkLqvq5r2u07MxToHgKybMQUuHpFggAvcwomi8nzx387hXJ7hB7KmTuxfOlXkbV9Gy+9vrfcJ-HcdG8O-pJgGjuT0kzvMpv6AGpw1jW9uFk75zh8Cma++80dB7B-ABeO7pz90SeN8j3RUDrNuQgXgrPeFc4B1zAeA2uICZjU1gCAiB8CEE5wbhZFuX5-4YxppfbuQVe6WH7g-PyV8R44LHlUCeC0N4u1nm7dGMDdZLwJnjI230N6fEjjvU6wDPgH2EHQY+5A0TQLCJgpy2Db54Pvr5N4T8Sq93eggIRmNP6-W-r-dBdCZHHS4dmUBiC9HwKgRAS22sYBwP0eYuuyDD6oI8sEYxOssHX1HlYfBUiiE3zsmQ8QyjuRUM5lANGcB7H0OkMww2y8WG-TYVvSWHC4baJ4fAPh64BFnyMagTWJjHHEPEa4x+CgX7yLsRkq2MAfERxiWtH+f8FHpMydbIBLM+BSSAA)

```typescript
export namespace Control {
  // PassThrough
  export class TControl<Type extends Type.TSchema> extends Type.Base<Static<Type>> {
    constructor(private readonly type: Type) { super() }
    public override Check(value: unknown): value is Type.Static<Type> {
      return Value.Check(this.type, value)
    }
    public override Clean(value: unknown): unknown {
      return Value.Clean(this.type, value)
    }
    public override Convert(value: unknown): unknown {
      return Value.Convert(this.type, value)
    }
    public override Create(): Type.Static<Type> {
      return Value.Create(this.type)
    }
    public override Default(value: unknown): unknown {
      return Value.Default(this.type, value)
    }
    public override Errors(value: unknown): object[] {
      return Value.Errors(this.type, value)
    }
  }
  // NoClean
  export class TNoClean<Type extends Type.TSchema> extends TControl<Type> {
    public override Clean(value: unknown): unknown {
      return value
    }
  }
  export function NoClean<Type extends Type.TSchema>(type: Type): TNoClean<Type> {
    return new TNoClean(type)
  }
  // NoConvert
  export class TNoConvert<Type extends Type.TSchema> extends TControl<Type> {
    public override Convert(value: unknown): unknown {
      return value
    }
  }
  export function NoConvert<Type extends Type.TSchema>(type: Type): TNoConvert<Type> {
    return new TNoConvert(type)
  }
  // NoDefault
  export class TNoDefault<Type extends Type.TSchema> extends TControl<Type> {
    public override Default(value: unknown): unknown {
      return value
    }
  }
  export function NoDefault<Type extends Type.TSchema>(type: Type): TNoDefault<Type> {
    return new TNoDefault(type)
  }
}
```

This approach seems quite sound. If this solves your original use case, I may actually look into this design more closely and possibly even consider inclusion into TB at a later time.

Let me know!
S

webstrand:
It looks like Base is a great way to customize and extend Convert behavior, though when Convert calls `Base#convert` it loses context information, and is an opaque type, so it's not able to use refs internally. Also it'd be nice if `FromType` could be exposed, it'd make inheriting and extending its behavior easier.

I'm no longer confident that either solution, keyword or `Base`, are a good way to work around the `Union` behavior of `Convert` and `Check`, though.

sinclairzx81:
> It looks like Base is a great way to customize and extend Convert behavior, though when Convert calls Base#convert it loses context information, and is an opaque type, so it's not able to use refs internally.

Yeah, the Context information is lost on Base as it was mostly written to be a self contained type (such as  Date or Uint8Array) and not necessarily one written to intercept and control Value.* referencing behaviors (the provided Control example was just something that fell out of the current implementation). 

This said, I had a quick look at what it would take to add the Context in there, but I think it needs more thought put into it as the Base type is currently serving as a Standard Schema integration point, and a Base type for compiled types, and TB referential types would be meaningless to remote embedded types (Zod, etc), and introducing a Context in a non-breaking way would require diverging from the signature / order pattern of (`context, type)`). There's a few things to think about, but it is tempting to include a Context somehow, but probably inline with a broader effort to implement some `Value.Process` design mentioned earlier.

With regards to type opacity though. Inference is explicit with Base, you can specify anything to infer.

```typescript
export class Foo 
  extends Type.Base<{ foo: 123 }> { }      // infers { foo: 123 }

export class Bar 
  extends Type.Base<Uint8Array> { }        // infers Uint8Array

export class Baz<Type extends TSchema> 
  extends Type.Base<Static<Type>> { }      // infers as Static<Type>
```

---

> I'm no longer confident that either solution, keyword or Base, are a good way to work around the Union behavior of Convert and Check, though.

Check should be correct as it is tested against the [official test suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite) (around 8K tests spanning several drafts), however the Convert logic could be investigated further.

#### Workaround 1: Swap order of A and B, ensuring the broadest form is first.

```typescript
const A = Type.Object({ a: Type.String() })
const B = Type.Object({ a: Type.String(), b: Type.String() })
const U = Type.Union([B, A])                            // swap
console.log(Value.Decode(U, { a: "x", b: "x" }))        // { a: "x", b: "x" }
```

#### Workaround 2: Add a descriminator.

```typescript
const A = Type.Object({ type: Type.Literal('A'), a: Type.String() })
const B = Type.Object({ type: Type.Literal('B'), a: Type.String(), b: Type.String() })
const U = Type.Union([A, B])
console.log(Value.Decode(U, { type: 'B', a: "x", b: "x" }))     // { type: 'B', a: "x", b: "x" }
```
#### Workaround 3:  ExclusiveOR (Value.* does not support `oneOf` traversal)

```typescript
const A = Type.Object({ a: Type.String() })
const B = Type.Object({ a: Type.String(), b: Type.String() })
const U = { oneOf: [A, B] } as const
console.log(Value.Decode(U, { type: 'B', a: "x", b: "x" }))     // "must match exactly one schema in oneOf"
```

Value.* logic related to Union has been a real pain point to get right. TypeBox needs to be able to return the same data irrespective of sub schema ordering, however achieving this has been exceptionally difficult and requires a dedicated effort to delve into logical type theory as it relates to the materialization and/or processing of values. V1 solves some of this for Intesection (And) but Union (Or) continues to prove elusive. 

---

> Also it'd be nice if FromType could be exposed, it'd make inheriting and extending its behavior easier.

Ah yeah, I may look into this, A similar thing was done for Codecs to let developers have at the raw Encode/Decode callback system.

```typescript
import { DecodeUnsafe, EncodeUnsafe } from 'typebox/value'
```

But for now,  I want to keep Value.* internals "mostly" private until a more formal plan can be devised to overhaul the namespace in support of Value.Process and JIT compilation (which means i need to keep things locked down to avoid heavy breaking API changes)

---

Will close out this one for now as I can't really take on this PR based on all that's been discussed, but hopefully all this sheds light into some of the more internal design considerations going on for the library. I would be happy to continue things more via discussion thread tho.

I will say, the Value.* module is written to be decoupled (mostly) from Type.*, and is largely in support of dropping in an entirely new Value.* namespace at some point in the future. The introduction of `Pipeline` was also to provide ways for users to take functions from `Value.*` and combine them with their own functions (including copying and pasting the current Value.* functions into local projects and modifying functionality as required). The current functions "are what they are", making the best of some really difficult design / architectural challenges relating to Json Schema, but I do stand by the view that TB should limit / restrict the construction of non-standard keywords (like `~convert`) as this promotes new Value.* modules being dropped in and doesn't imply keyword behaviors of other Json Schema validators (such as Ajv). TB supports user defined custom keywords (with effort), just not embedded into the library itself.

Thanks for the PR Draft and the chat! :)
S

webstrand:
Er I made a major typo in my last comment: it's the the `Union` behavior of `Convert` and ~~Check~~ **`Clean`** that have ambiguous behavior.

`Check` as you say is no problem at all

Issue #1364, opened 2025-09-26
Please add .md docs for LLM


caracal7:

sinclairzx81:
@caracal7 Hello!

What would this entail? Do you have a reference to some prior art or a documentation link?

caracal7:
@sinclairzx81 👋

You just need to put all the documentation into a single Markdown file and share the link on the official website and in the repository. That's the minimum and it should be sufficient. 

You can see documentation examples here: 
[https://svelte.dev/docs/llms](https://svelte.dev/docs/llms), and there's also a link to a document describing the proper formatting. 

This will allow users of Cursor and similar editors to use all the latest features of TypeBox. 

Thanks for the amazing library! 💪

jaydenfyi:
@caracal7 The entire documentation is in the [readme file](https://github.com/sinclairzx81/typebox/blob/main/readme.md)

I'd recommend using the Context7 MCP Server https://context7.com/sinclairzx81/typebox

Not-Jayden:
*EDIT: I'm clearly out of the loop, did not realise there was a docs site now 🤦 

Maybe a good alternative for the time being but please ignore me.

caracal7:
> [@caracal7](https://github.com/caracal7) The entire documentation is in the [readme file](https://github.com/sinclairzx81/typebox/blob/main/readme.md)
> 
> I'd recommend using the Context7 MCP Server https://context7.com/sinclairzx81/typebox

That's exactly what I need.
Thanks a lot)

@sinclairzx81 You can just provide the link to Context7 [typebox | context7](https://context7.com/sinclairzx81/typebox/llms.txt)

sinclairzx81:
@caracal7 Hiya, I might close this one out.

> @sinclairzx81 You can just provide the link to Context7 [typebox | context7](https://context7.com/sinclairzx81/typebox/llms.txt)

Not keen to add links to remote LLM / context sites at this stage.

I will add LLM context information to TB at some point in the future, but maybe once the ecosystem establishes some standards, conventions and guidelines. As it stands, the TypeBox API is very large (and may increase again early next year), I will revisit things next year. 

[...]

PR #1399, opened 2025-10-14
feat: Add `llms.txt` to docs

jaydenfyi:
Relates to #1364

### DIFF
--- a/docs/llms.txt
+++ b/docs/llms.txt
@@ -1,0 +1,4478 @@
+# Source: docs/overview/overview.md
+
+# TypeBox
+
+A Runtime Type System for JavaScript
+
+## Overview
+
+TypeBox is a runtime type system that creates in-memory Json Schema objects that infer as TypeScript types. The schematics produced by this library are designed to match the static type checking rules of the TypeScript compiler. TypeBox offers a unified type system that can be statically checked by TypeScript and validated at runtime using standard Json Schema.
+
+This library is designed to allow Json Schema to compose similar to how types compose within TypeScript's type system. It can be used as a simple tool to build up complex schematics or integrated into REST and RPC services to help validate data received over the wire.
+
+## Features
+
+## Type
+
+The Type namespace creates Json Schema.
+
+```typescript
+import Type from 'typebox'
+
+const T = Type.Object({                             // const T = {
+  x: Type.Number(),                                 //   type: 'object',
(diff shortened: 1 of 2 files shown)


sinclairzx81:
Hi, thanks for the PR!

Things seem ok here (afaik  :D), but just a couple of general questions first.

1) Does the file extension need to be `.txt`? (can it be `.md`?)
2) The generated file is 4000+ lines long, just wondering about min/max token counts for LLM's to ingest all this content (I note that some projects provide a summarized view, would that make sense here?).
3) Is there a way to test this? (Note: I'm not a LLM/AI user so will need some guidance here)
4) Should each sub module `/value` `/compile`, `/format`, etc have it's own LLM file? (documentation splitting - see 5)
5) if splitting documentation across multiple files, do LLM systems support loading referenced documentation in multiple files? (I see the output currently has `# Source: docs/overview/overview.md`, but does it support `# Reference: ...` )?

---

Thanks for taking the time to look into this. If you can provide some info on the above aspects (especially the testing / splitting aspect) that would be awesome. 

As mentioned, I'm not really a user of AI systems, so I'm not sure exactly what the expectation would be or what best practice is.. but if I am to include anything, it would probably need to be supported by all the major LLM model vendors (Claude, OpenAI, Gemini, LLaMA, Orca, etc). If you can provide a bit more information to help me get my bearings on this stuff, I will do a quick review and can probably just merge this stuff through. 

With regards to the PR itself, the only other thing I can think of is that it might be better to create a separate module specific to generating the LLM document rather than updating the code to generate the Website (as I may be overhauling that code at some point with updates specific to the website). It might be better to create a new task specific for LLM generation, what do you think?

Cheers!
S

sinclairzx81:
@jaydenfyi Heya, thanks again for this PR.

I think at this stage, what I might end up doing is just manually crafting a LLM document, just something that includes all the functions provided by TypeBox and shows a single line example of how to use each.

The issue with just bundling all the documentation into a single file is that it's HUGE! (and likely going to increase once the new schema compiler gets documented). My understanding is that most LLM's will have difficulty with "too much context" and I expect 4000+ (soon to be 5000+) is just going to be too much for most.

Will close this one out for now, but if you have any follow thoughts on how example generation could be automated (without documentation concat) I would be keen to hear. I am considering writing a small compiler API tool for this, but will see how things go.

Thanks again
All the best
S

PR #1403, opened 2025-10-15
resolves #1402: implement toJSON for string types to handle Regex patterns

### LINKED ISSUE #1402: Shouldn't string regexes serialize to JSONSchema pattern strings?
Consider this type:

```typescript
export const componentIdSchema = Type.String({
  pattern: /^@component\//,
  description:
    'The unique identifier for the component, e.g. @component/button. Must start with "@component/".',
});
```

When this is serialized with `JSON.stringify`, it becomes:

```json
{
  "type":"string",
  "pattern":{},
  "description":"The unique identifier for the component, e.g. @component/button. Must start with \\"@component/\\"."
}
```

I _expected_ to get:

```json
{
  "type":"string",
  "pattern": "/^@component\//",
  "description":"The unique identifier for the component, e.g. @component/button. Must start with \\"@component/\\"."
}
```

FWIW `String(componentIdSchema) === '[object Object]'`, which is entirely reasonable.

I would characterize this as probably a feature request, rather than a bug, but I would note that this behavior differed from what I expected to happen... readers can decide whether or not my expectation was reasonable :)

dancrumb:
I built locally and ran with the type in the associated PR and got:

```json
{
  "type":"string",
  "pattern":"^@component\\/",
  "description":"The unique identifier for the component, e.g. @component/button. Must start with \"@component/\"."
}
```

which is what I would have expected

### DIFF
--- a/src/type/types/string.ts
+++ b/src/type/types/string.ts
@@ -29,39 +29,31 @@ THE SOFTWARE.
 // deno-lint-ignore-file
 // deno-fmt-ignore-file
 
-import { Memory } from "../../system/memory/index.ts";
-import {
-  type TSchema,
-  type TSchemaOptions,
-  type TStringOptions,
-  IsKind,
-} from "./schema.ts";
+import { Memory } from '../../system/memory/index.ts'
+import { type TSchema, type TSchemaOptions, type TStringOptions, IsKind } from './schema.ts'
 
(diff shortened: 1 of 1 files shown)


dancrumb:
Having seen your response to my issue, I no longer believe that this is a useful fix

PR #1427, opened 2025-10-30
fix: add missing iso-date-time

melroy89:
You are missing `iso-date-time`. 

And yes, its **NOT** the same as `date-time`, see also the official Avj docs: https://ajv.js.org/packages/ajv-formats.html#formats

> iso-date-time: date-time with optional time-zone.

I was specifically using `iso-date-time` format in my old JSON schema.. (bleh):

```ts
  occurred_date: {
    type: 'string',
    format: 'iso-date-time',
    description: 'Format: YYYY-MM-DDThh:mm:ss'
  }
```

Which I now want to convert to... Typebox 🥳 .. (yeah)

You might also want to update https://github.com/sinclairzx81/typebox/blob/main/src/format/date-time.ts ? Or create a dedicated `iso-date-time.ts` file...??

### DIFF
--- a/src/type/types/schema.ts
+++ b/src/type/types/schema.ts
@@ -228,6 +228,7 @@ export type TFormat =
   | 'ipv6'
   | 'iri-reference'
   | 'iri'
+  | 'iso-date-time'
   | 'json-pointer-uri-fragment'
   | 'json-pointer'
   | 'json-string'
@@ -255,6 +256,7 @@ export interface TStringOptions extends TSchemaOptions {
    * - `ipv4` – IPv4 address.
    * - `ipv6` – IPv6 address.
    * - `iri` / `iri-reference` – Internationalized Resource Identifier.
+   * - `iso-date-time` – [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) date-time format with optional timezone.
    * - `json-pointer` / `json-pointer-uri-fragment` – JSON Pointer format.
    * - `json-string` – String containing valid JSON.
    * - `regex` – Regular expression syntax.


sinclairzx81:
@melroy89 Hi, thanks for the PR.

Unfortunately, I can't include additional formats just yet as I'm not sure how I want to approach adding formats above those defined by the JSON Schema specification. The spec formats implemented by TB are listed at the URL below, and where TB only advertises support for these specifically.

https://json-schema.org/understanding-json-schema/reference/type#built-in-formats

---

> You might also want to update https://github.com/sinclairzx81/typebox/blob/main/src/format/date-time.ts ? Or create a dedicated iso-date-time.ts file...??

Yes! If taking on an additional format, the format would need a RFC specification (or similar) and there would need to be a validator written to the `src/format/*` sub module as well as a bunch of tests that assert it's validating correctly. 

The current suite of formats were only really considered possible in TB due to the official [JSON Schema test suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite) providing spec compliant test cases for each format noted in the JSON Schema spec. The Format sub module was only added in 1.0.

For additional formats to be considered

1) They would need a RFC / IETF (or similar) specification 
2) They would need a test suite written that conforms to the spec.
3) There would need to be some observed usage of the format (databases, applications, etc)

---

At this stage, I am in the process of finalizing JSON Schema 2020-12 spec compliance (hopefully to be ready by the end of this year), at which point I'll consider how best to take on additional formats.

In mean time, additional formats can be registered external to TB using the Format registry.

https://sinclairzx81.github.io/typebox/#/docs/format/1_registry

And the `TFormat` union is open to external extension via `({} & string)` 

Thanks again, but will close this one out for now. Will take another look at additional formats in the new year.
Cheers!
S

melroy89:
That is indeed unfortunate, but understandable.

I can confirm that using the format string "iso-date-time" does work luckily within my Typebox setup. 
Despite it's not part of the TFormat options.. It seems that tsc is just fine about it? 

It's working with my setup now with Fastify (and also Swagger UI does shows it correctly).

sinclairzx81:
@melroy89 Hiya,

> I can confirm that using the format string "iso-date-time" does work luckily within my Typebox setup.
Despite it's not part of the TFormat options.. It seems that tsc is just fine about it?

Yes, Formats are extensible using the `'A' | 'B' | 'C' | ({} & string)` technique where `'A' | 'B' | 'C'` is observable by the type system (so you get auto complete), but where `({} & string)` allows for arbitrary string assignment.

```typescript
type T = 'A' | 'B' | 'C' | ({} & string)

function Test(value: T) {}

Test('A')
Test('B')
Test('C')
Test('D')  // fine because of ({} & string)
```

Technically the `format` keyword can be assigned any string value, but TB provides the `TFormat` just to give a quick indication which formats are built in via auto complete.

I will look at formats again in the new year as I do think there is potential to include more, I just need to finish the 2020-12 spec first then figure out how best to provide them (initial thoughts are via external library, but we'll see)
Cheers!
S

Issue #1429, opened 2025-10-30
Use of Static in function overload causes ts(2589) Type instantiation is excessively deep and possibly infinite


holvi-sebastian:
This one has me scratching my head rather vigorously...

If I use `Static` in an overload signature, I get the dreaded `ts(2589) Type instantiation is excessively deep and possibly infinite` on the `myFn` overload:

```ts
//         |----- here!
function myFn<T extends TSchema>(schema: T): Static<T>
function myFn(schema: TSchema): unknown {
  return {} as any
}
```

Here's a [demo in TS Playground](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgFQJ5gKYBoUGUDGAFhiAIY64ykzD5wC+cAZlBCHAOQzoYBGEADw4AoYUwCuAO3w0IkuCFQAxSQB5kcDAJgZJAEwDOeIiVIA+ABQGTZAFwoAlPcrVa6s2KkzgchcslWNqT2yATEZE5wUgDWkhAA7vIIwnBwUBgw4lBJjKRGpJKowvRAA)

The error only appears on overloads, seemingly always the first one, if it or any other overload uses `Static`.

sinclairzx81:
@holvi-sebastian Hi,

This is a known TypeScript problem, you can use the following to work around it.

[...]

Issue #1428, opened 2025-10-30
TSchema is empty and matches any type.


holvi-sebastian:
Currently, `TSchema` is defined as 
```
export interface TSchema { }
```

As an empty object, this type is equivalent to basically everything, which leads to unexpected results. For example, when using the `Type.IsSchema` guard, the negated type is `never`.

Example:

```ts
const x = 'blah' as const
    
if(Type.IsSchema(x)) {
  expectTypeOf(x).toBeString()
}
else {
  expectTypeOf(x).toBeNever()
}
```

Would it be a good idea to add an identifier property to `TSchema` to ensure that a `typebox` schema can be distinguished from other types?

Maybe this would make sense:
```ts
export interface TSchema {
     ['~kind']: Kind;
}
```

sinclairzx81:
@holvi-sebastian Hi, 

> Would it be a good idea to add an identifier property to TSchema to ensure that a typebox schema can be distinguished from other types?

This is actually by design. In version 1.0, all objects are considered valid TypeBox types. This is one of the more significant changes in 1.0 where TypeBox is moving towards full JSON Schema compliance .... and where empty objects are considered `unknown`.

It's not currently documented (yet), but TypeBox 1.0 is effectively 95% JSON Schema 2020-12 compliant and where the library has been heavily re-architected to be a viable alternative to Ajv for direct JSON Schema validation (without Type.*). The decision to consider all objects as valid TSchema falls inline with how JSON Schema reasonable about validation constraints, and as a result, V1 makes some fairly significant changes to how types are reasoned about in the library.

Ill try explain some of the design below.

---

## Evaluation Types

V1 has a more sophisticated inference system to V0 based on type evaluation. Rather than relying on specific "fixed" JSON Schema structures (as was the case in V0), V1 is moving towards a logical evaluation strategy inline with how JSON Schema validates data constraints. 

It's a little difficult to communicate all the reasoning that's gone into V1 (it's mostly academia and it's not all that interesting), but the following is a general high level showing how V1 reasons about types, and where logical evaluation is used to narrow from `unknown` (empty schema) to a final type `{ x, y }`

[...]

PR #1430, opened 2025-10-31
Fix StaticType causes intermittent ts(2589) Type instantiation is excessively deep and possibly infinite

holvi-sebastian:
As described in: https://github.com/sinclairzx81/typebox/issues/1429
And: https://github.com/sinclairzx81/typebox/issues/1428

This PR adds an escape to `StaticType` allowing `typescript` to skip evaluation when `Type` is equivalent to `{}`. Without this, `typescript` is forced to exhaust all cases to arrive at `unknown` which, since `TSchema` is an empty object and equivalent to almost everything, causes infinite type expansion in many unexpected cases.


holvi-sebastian:
Set this to draft because although this does resolve `ts2589` and all `typebox` repo tests still pass, it does change `Static` semantics slightly.

For example, the following becomes an error:

```ts
const Schemas = {
  x: Type.Object({
    path: Type.String(),
  }),
}

type Schemas = typeof Schemas
type SchemasKey = keyof Schemas

//                               |---- Here!
type X<K extends SchemasKey> = Static<Schemas[K]>['path']
```

```
Type '"path"' cannot be used to index type 'StaticType<[], "Encode", {}, {}, { x: TObject<{ path: TString; }>; }[K]>'.
```

holvi-sebastian:
Wow, here be dragons... This really is quite finicky. 

I get the same `ts2589` if I change `Static` to work with `TKind` and drop support for raw JSON Schema obj, so it's not even related to that. It's just the size of the type, I guess?

I guess the only reason I gave it this much attention is because it's quite jarring to have a `ts2589` thrown in your face when you use `Static` in a seemingly completely reasonable way. I think we should avoid requiring special handling from devs where possible.

Do you have any ideas, or are we doomed to use your default `Result` approach?

sinclairzx81:
@holvi-sebastian Hiya,

> I guess the only reason I gave it this much attention is because it's quite jarring to have a ts2589 thrown in your face when you use Static in a seemingly completely reasonable way. I think we should avoid requiring special handling from devs where possible.

> Do you have any ideas, or are we doomed to use your default Result approach?

So, I had another look into this tonight and there was actually a issue here. The issue related to the way TB was handling Codec inference where the Codec destructure (infer) wasn't extracting the interior TSchema (it was just extracting another Codec leading to infinite Static)

The PR to fix can be found https://github.com/sinclairzx81/typebox/pull/1431/files and been published on 1.0.45.

[TypeScript Link Here](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzjAnmApnAyjAhjYAYwBpk1MAVLQgC3RF1IvLgF84AzKCEOAclQYARhAAefAFASOAVwB2hAhDlwQKAGJyAPBTjpRMdHIAmAZzhVa9XAD4AFKasMAXBYCUrnPiI6b0+YrAyqoacg5OuK6WdAwecPIA1nIQAO4qCBJwcFDoMDJQ6ey45rhyKBKsEoTKpvA5pjIANvAAvCGadswYAHTUUMBgMHYABplkGHAAknJgMq2IY1miAPyucjIgQuhQi3Aoq3Drm9u7AF4HR1s7WZVZgpgAsrhgvnBtGVlZObjGyo0ocAA2gBpODAFQJdAoCAcCwAXQAtAcKCC4XAAD4LT6fBhgDDGVwAVTx20IxXQWmBfmxtzYUjuLAA8nNZvMni9pqy-BJhm5usyYKy3BIgA) (you may need to refresh you browser cache to get the TSP pulling the latest)

```typescript
import { type Static, type TSchema, Type } from 'typebox'

function myFn<T extends TSchema>(schema: T): Static<T>
function myFn(schema: TSchema): unknown {
  return {} as any
}
const result = myFn(Type.Script(`
  type Input = {
    x?: number
    y?: number
    z?: number
  }
  type Map<T> = {
    readonly [K in keyof T]-?: T[K] | {
      mapped: Uppercase<K>
    }
  }

  type Output = Map<Input>

`).Output)
```

---

Thanks for pressing on with this btw, this issue probably wouldn't have been fixed if you hadn't pressed with the PR and follow up comments. I note that V0 suffered a similar issue (tho for a different reason) so I had resigned this to being TS overloaded signature behavior without giving much thought to it being an actual bug, mostly because the inference behaviors mirror that of V0.

Actually having this solved (in light of it being a bug) is actually a big win for TB, so yeah, actually really appreciated the follow ups! :)

---

Happy to discuss more here or to just close out this draft. If you want to test the 1.0.45 updates locally and let me know all is working, that would be good. I will leave this Draft open for a few days if you want to comment or suggest additional updates, or can just close out.

Thanks again!

martaver:
I'm glad to hear it helped :) I think `typebox` has a really unique take on the whole runtime types scene, and hope for its success! We're using it to define types we use in E2E tests over here, and use it hand-in-hand with `json-schema-faker`.

> The issue related to the way TB was handling Codec inference where the Codec destructure (infer) wasn't extracting the interior TSchema (it was just extracting another Codec leading to infinite Static)

I would actually love to know the process you used to diagnose this error, was it just a process of elimination, commenting out each case of `StaticType`?

sinclairzx81:
> I'm glad to hear it helped :) I think typebox has a really unique take on the whole runtime types scene, and hope for its success! We're using it to define types we use in E2E tests over here, and use it hand-in-hand with json-schema-faker.

Ah cool. Glad to hear the library is helping out :)

> I would actually love to know the process you used to diagnose this error, was it just a process of elimination, commenting out each case of StaticType?

Yeah pretty much. For this specific issue, I only really decided to take a closer look based on this specific comment.

> It's just the size of the type, I guess?

... where I wouldn't have expected the size of the type to be an issue. I note that V1 uses a recursive parse / evaluation strategy over the V0 type level caching (formally the `static` property). At most, I was just trying to understand if I needed to return to V0 type level caching (due to the size of the type) or if there was something else going on. The Codec issue revealed itself after commenting the StaticType matchers and where I noticed your repro case was solved on commenting out the Codec arm.

---

There's a bit of an art-form to debugging type level programs of this size, but you were in the right place looking at StaticType. TB V1 was written to try make type level debugging a bit easier using a few techniques and patterns that have evolved over the years, but the degree in which it holds up over the course of V1 revisions will likely inform whatever V2 needs to be, as these things tend to go.

Will close this one out for now, but thanks again! This was a good find :)
All the best!
S

PR #1460, opened 2025-11-21
fix: Don't skip required undefined properties on object

orimay:
Currently, if we run `Value.Default` on an object, that has a required property, that is undefined by default, TypeBox ignores it. This PR fixes this bug

### DIFF
--- a/changelog/1.0.0-migration.md
+++ b/changelog/1.0.0-migration.md
@@ -84,16 +84,20 @@ The following implements Uint8Array using Type.Base.
 ```typescript
 import Type from 'typebox'
 
-// Definition
 export class TUint8Array extends Type.Base<Uint8Array> {
+  // required: Used by validation
   public override Check(value: unknown): value is Uint8Array {
     return value instanceof Uint8Array
   }
+  // required: Used by validation
   public override Errors(value: unknown): object[] {
-    return !this.Check(value) ? [{ message: 'Not a Uint8Array'}] : []
+    return !this.Check(value) ? [{ message: 'not a Uint8Array'}] : []
+  }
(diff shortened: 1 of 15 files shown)


sinclairzx81:
@orimay Hi, thanks for the PR

I think this one is ok, but will want to include some tests as this change would implicate Default return values. The test module is at the link below, so will want to drop in a few tests here (just add them to the bottom of the test module and include link back to this PR via comment)

https://github.com/sinclairzx81/typebox/blob/main/test/typebox/runtime/value/default/object.ts

Also, looks like there's been some merge conflicts introduced since the PR commits (there's been a few rebases on the main branch related to a couple of documentation passes). Would be good to sync things up such that only modified files related to this PR show up, just helps narrow things to only files that have changed  Apologies for the upstream changes.

Thanks again!

orimay:
> @orimay Hi, thanks for the PR
> 
> I think this one is ok, but will want to include some tests as this change would implicate Default return values. The test module is at the link below, so will want to drop in a few tests here (just add them to the bottom of the test module and include link back to this PR via comment)
> 
> https://github.com/sinclairzx81/typebox/blob/main/test/typebox/runtime/value/default/object.ts
> 
> Also, looks like there's been some merge conflicts introduced since the PR commits (there's been a few rebases on the main branch related to a couple of documentation passes). Would be good to sync things up such that only modified files related to this PR show up, just helps narrow things to only files that have changed Apologies for the upstream changes.
> 
> Thanks again!

@sinclairzx81, sorry, closed this once unintentionally, by force-pushing your upstream to my branch, but here's the fix with tests now: https://github.com/sinclairzx81/typebox/pull/1463

PR #1463, opened 2025-11-24
fix: Don't skip required undefined properties on object

orimay:
Continuation of the https://github.com/sinclairzx81/typebox/pull/1460

### DIFF
--- a/src/value/default/from-object.ts
+++ b/src/value/default/from-object.ts
@@ -41,7 +41,7 @@ export function FromObject(context: TProperties, type: TObject, value: unknown):
   const knownPropertyKeys = Guard.Keys(type.properties)
   // Properties
   for (const key of knownPropertyKeys) {
-    // Resolve: Value from properties.
+    // Resolve Value for Property
     const propertyValue = FromType(context, type.properties[key], value[key])
 
     // Ambiguious Undefined: If the value is undefined, the type is optional there's no default. ignore.


sinclairzx81:
@orimay Hi!

Thanks for keeping this one going.

---

So, just been going through the tests. I note some of the logic around optional / undefined property handling can be quite complex (the Value.* API has been notoriously tricky to get right), but will want to try keep current value generation behaviors as they are while still bringing in these updates.

From a PR standpoint, will want ensure additive tests only (23, 24), but retain the existing tests as they are.

### Tests

---

This one is good, we would expect a default value of `undefined` to be generated from the annotation.

```typescript
Test('Should Default 23', () => {
  const X = Type.Object({ x: Type.Undefined({ default: undefined }) })
  const R = Value.Default(X, {})
  Assert.IsEqual(R, { x: undefined })
})
```
---

This one just needs an update to correctly check optional / variance handling for Undefined
```typescript
Test('Should Default 24', () => {
  // const X = Type.Object({ x: Type.Optional({ default: undefined }) }) // <--- no embedded Undefined
  const X = Type.Object({ x: Type.Optional(Type.Undefined({ default: undefined })) })
  const R = Value.Default(X, {})
  Assert.IsEqual(R, {})
})
```

---

Unfortunately, will need to revert these as updates here would signal a breaking change. I think the property generation logic should be something like ...

> If the `default` annotation is NOT present OR the property is NOT `TUndefined` then no key should be generated

But see how you go, mostly just want to keep these existing tests as is.

```typescript
Test('Should Default 7', () => {
  const T = Type.Object({
    x: Type.Number(),
    y: Type.Number()
  })
  const R = Value.Default(T, {})
  Assert.IsEqual(R, { x: undefined, y: undefined }) // <-- unexpected key generation
})
Test('Should Default 8', () => {
  const T = Type.Object({
    x: Type.Number({ default: 1 }),
    y: Type.Number()
  })
  const R = Value.Default(T, {})
  Assert.IsEqual(R, { x: 1, y: undefined })  // <-- unexpected key generation
})
Test('Should Default 9', () => {
  const T = Type.Object({
    x: Type.Number({ default: 1 }),
    y: Type.Number()
  })
  const R = Value.Default(T, { x: 3 })
  Assert.IsEqual(R, { x: 3, y: undefined })  // <-- unexpected key generation
})
```

This would apply to the other updated tests too. Additive tests only.

---

The code updates look ok, but just need some additional guards to check specifically for properties of `TUndefined`. This should enable the other tests to pass.

Let me know how you go, if you have questions, feel free to ping!
Cheers
S

orimay:
Made it only populate the key what there's a default specified and the key is not optional

orimay:
@sinclairzx81 Hmm, I test these changes locally (as a Bun patch) before you merge them, and must point out that not having non-optional fields defaulted to undefined no present is a real pain. I have an `id` property, that needs to be there. I define it in schema like this:
```ts
const schemaTask = Type.Object({
  id: Type.Union([Type.Undefined(), schemaIdTask], { default: undefined }),
});
```
But when I run the following, `id` is not there, and the further logic (Proxy that depends on keys present in the object) fails:
```
Value.Default(schemaTask, {});
```

Unexpected and upsetting behavior, I think I will create an util that goes over schema and creates non-optional fields. Would you like me to make any more changes to the logic/tests, or is this PR good to go?

sinclairzx81:
@orimay Hi, thanks for the updates!

---

> Would you like me to make any more changes to the logic/tests, or is this PR good to go?

I think the PR looks ok!! The only other suggestion I can think of would be to use


```typescript
Guard.HasPropertyKey(type.properties[key], 'default') 

// instead of 

'default' in type.properties[key]
```
But it's optional, you can leave it as is for now (I can update on review), but if you want to update, that's fine too :)

> Note: TB puts as many checks behind Guards as there is some potential to move TB to other languages (Python/Ruby/Lua) and where the Guards help make the operations and checks performed on values a bit more abstract (for example, you could imagine writing a similar Guard in another language).

---

> But when I run the following, id is not there, and the further logic (Proxy that depends on keys present in the object) fails:

Ah, it might be a consequence of Undefined. Unions (and Intersect) have been a real issue for Value.* functions (logical types require a degree of special handling), was this issue related to the Undefined update, or is it existing behavior?

---

I will need sit down with the updates locally ran run a few tests (hopefully tomorrow) and can likely merge through soon after (I may drop in a couple of other tests). I had meant to get to things today, I've been a bit stretched for time unfortunately. But everything looks ok to me :)

orimay:
@sinclairzx81 Done!

The `undefined` behavior is the one that currently happens. It is particularly what you asked me to revert above. I was expecting `undefined` be there mandatory, unless the field is optional. Because you may either have a field with `undefined` value present or have no field at all. And if schema states that there's a field with `undefined` value, it must be there, unless the field is optional. Otherwise, the schema doesn't reflect the shape. I would consider it a bug, but I understand that fixing it would be a breaking change. Perhaps, something to consider for a minor version bump?

Here's an outline of what I'm talking about:
```ts
type A = {
  a: undefined; // this field is required and must be present on object, having a value of `undefined`
};

type B = {
  b?: undefined; // this field is optional and may not be present on object at all or have a value of `undefined`
};
```
So if I describe type `A`, the field `a` absolutely must end up in the object after calling `Value.Default` on it, while type `B` is allowed to ignore it on the `Value.Default` call

sinclairzx81:
@orimay Hi,

Everything looks good here.

> The undefined behavior is the one that currently happens. It is particularly what you asked me to revert above. I was expecting undefined be there mandatory, unless the field is optional. Because you may either have a field with undefined value present or have no field at all. And if schema states that there's a field with undefined value, it must be there, unless the field is optional. Otherwise, the schema doesn't reflect the shape. I would consider it a bug, but I understand that fixing it would be a breaking change. Perhaps, something to consider for a minor version bump?

Ah, undefined handling is a real pain unfortunately. TypeBox tries to model for the `exactOptionalPropertyTypes: true` setting in TS, but often times runtime behavior warrants a `exactOptionalPropertyTypes: false` and I'm usually lost trying to reconcile Value.* functions in accordance with either interpretation. 

TypeBox has it's own `exactOptionalPropertyTypes` setting to align with whatever TS is configured for (so it's technically possible to return varying behaviors behind this setting - without version bump), but I think in this case, the runtime behaviors is ok.

https://sinclairzx81.github.io/typebox/#/docs/system/1_settings

---

I have added a few additional tests just to assert the current behavior for the `schemaTask`, but I am open to changing it if you see a better alternative. As mentioned, a lot of the Value behaviors can be quite "interpretive", but interpretations can change if presented with better ones. Can review in a subsequent PR! :D

Sorry it's taken a bit to get this one through, have been quite pressed of late, but all things look good here, will merge through and publish out on 1.0.61.

Thanks again!
S

sinclairzx81:
@orimay Updates published on 1.0.61. Thanks again

orimay:
Awesome, thank you!

Issue #1466, opened 2025-11-27
Type.Intersect runs codec twice over the same property


orimay:
If we have an intersection of types, if the same property appears multiple times, the codec will run over it as many times as it appears instead of only being run once. Took me three days to figure out what the issue is :D

```ts
import { Type } from 'typebox'
import { Value } from 'typebox/value'

const schemaId =
  Type.Codec(
    Type.Object({
      table: Type.String(),
      id: Type.String()
    })
  )
    .Decode(encoded => (`${encoded.table}:${encoded.id}`))
    .Encode(decoded => {
      const [table, id] = decoded.split(':') as [string, string]; // once runs good, builds an object, another time `decoded` is that object
      return {
        table,
        id,
      };
    })

const schemaU = Type.Intersect([
  Type.Object({ id: schemaId }),
  Type.Object({ id: schemaId }),
]);

const value = Value.Encode(schemaU, { id: 't:x' }); // decoded.split is not a function or its return value is not iterable
```

Please see here in action:
[...]

PR #1471, opened 2025-11-28
fix: Intersect codec multirun

orimay:
Fixes https://github.com/sinclairzx81/typebox/issues/1466

### DIFF
--- a/src/value/codec/from-intersect.ts
+++ b/src/value/codec/from-intersect.ts
@@ -29,17 +29,61 @@ THE SOFTWARE.
 // deno-fmt-ignore-file
 
 import { Guard } from '../../guard/index.ts'
-import { type TIntersect, type TProperties } from '../../type/index.ts'
+import { type TIntersect, type TProperties, type TSchema, IsObject, IsOptional } from '../../type/index.ts'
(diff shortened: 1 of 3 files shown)


orimay:
@sinclairzx81 Not sure if you will find my approach the best (since I only let a single schema modify every key). Another approach I would consider is to let all the keys be processed my all the schemas and then throw if there's a difference in outcome. Not sure if it would be a better DX though.

sinclairzx81:
@orimay Hi, thanks for this PR and again, apologies for the delay.

This issue should be resolved via PR https://github.com/sinclairzx81/typebox/pull/1566 and have put some additional information on the associated issue here https://github.com/sinclairzx81/typebox/issues/1466#issuecomment-4197979339.

Really sorry this took so long to get to, I had been meaning to sit down and go through some of the logic here much sooner, but the Codec system in TB has had a fairly low priority to the stabilization work around the new 1.x infrastructure. But found some time today.

Will close this one out for now, but happy to discuss more on the issue thread if you have any follow up questions on the https://github.com/sinclairzx81/typebox/pull/1566  implementation.

All the best!
S

Issue #1470, opened 2025-11-28
`Type.Evaluate` doesn't respect `Type.Base` during `Type.Intersect`


orimay:
When I try to intersect a type extended from base and then evaluate it, it turns into `never`

```ts
import { Type, type StaticDecode } from 'typebox';

class TId extends Type.Base<string> {}

const schemaNonEvaluated = Type.Intersect([
  new TId(),
  new TId(),
]);
type TNonEvaluated = StaticDecode<typeof schemaNonEvaluated>;

/**
 *  const schemaNonEvaluated: Type.TIntersect<[TId, TId]>;
 *  type TNonEvaluated = string;
 */

const schemaEvaluated = Type.Evaluate(schemaNonEvaluated);
type TEvaluated = StaticDecode<typeof schemaEvaluated>;

/**
 *  const schemaEvaluated: Type.TNever;
 *  type TEvaluated = never;
 *  
 *  instead of
 *  
 *  const schemaEvaluated: TId;
 *  type TEvaluated = string;
 */
```

[...]

PR #1480, opened 2025-12-08
fix: EType.Evaluate respects Type.Base during Type.Intersect

orimay:
Fixes https://github.com/sinclairzx81/typebox/issues/1470

### DIFF
--- a/src/system/memory/discard.ts
+++ b/src/system/memory/discard.ts
@@ -30,14 +30,16 @@ THE SOFTWARE.
 
 import { Metrics } from "./metrics.ts"
 import { Clone } from './clone.ts'
+import { IsBase } from '../../type/types/base.ts'
 
 type ObjectLike = Record<PropertyKey, any>
 
 /** Discards multiple property keys from the given object value */
 export function Discard(value: ObjectLike, propertyKeys: PropertyKey[]): ObjectLike {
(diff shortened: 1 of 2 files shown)


sinclairzx81:
@orimay Hiya,

Will close this one out as Base has been flagged for deprecation (so won't have any additional infrastructure written for it). The recommended alternative moving toward will be Refine/Unsafe. As follows.

```typescript
// -------------------------------------------
// Base (Deprecated)
// -------------------------------------------
export class TDate extends Type.Base<File> {
  public override Check(value: unknown): value is File {
    return value instanceof globalThis.Date;
  }
  public override Errors(value: unknown): object[] {
    return this.Check(value) ? [] : [{ message: "must be Date" }];
  }
  public override Clone(): TDate {
    return new TDate();
  }
}
// -------------------------------------------
// Refine/Unsafe (Modern)
// -------------------------------------------
const DateType = () => Type.Refine(Type.Unsafe<Date>({ type: 'Date' }), 
  value => value instanceof globalThis.Date,
   'must be Date')
```

Thanks for the PR (and sorry for the delay on this!), Open to discussions and feedback on the Refine/Unsafe approach. Any missing aspects or features will be considered in the next significant release.

All the best!
S

PR #1491, opened 2025-12-23
Version 1.0.67

sinclairzx81:
This PR is a continuation of 1.0.66 that adds a dedicated StringGuard module for optimized Grapheme Cluster counting. The module provides fast `IsMinLength` and `IsMaxLength` checking with fallback to the more expensive Grapheme Cluster counting when encountering a clustering code point.


qeleb:
@sinclairzx81 This PR breaks validation for Strings with `minLength` 0. The Decode below works as expected in version `1.0.66`, but fails in all versions after (1.0.67 - 1.0.78).
```
const schema = Type.String({ minLength: 0, maxLength: 9 });
Value.Decode(schema, '');
```

sinclairzx81:
@qeleb Hi, thanks for letting me know! This is a good catch.

Have published a fix for this on 1.0.79. Associated PR here https://github.com/sinclairzx81/typebox/pull/1518

Cheers!

Issue #1522, opened 2026-01-24
Instantiate drops options on Ref


p2004a:
When I create a type like:

```ts
const schema = Type.Record(
    Type.String(),
    Type.Object({
        prop: Type.Ref("refTarget", { $comment: "about what this field means here" }),
    })
)
```

I would expect to see `$comment` present when I simply do `JSON.stringify(schema)`. Just like it's present without `Type.Record`. We use it as documentation inside of the JSON Schema files.

It looks that `Type.Record` does some type instantiation and https://github.com/sinclairzx81/typebox/blob/2d69ad29aa1881bc2acdc12ed8910b6a298bc799/src/type/engine/instantiate.ts#L377 and `RefInstantiate` don't pass around schema options attached to original `Ref`.

Is there some reason why the original options on `Ref` are not passed around?

It is a change in behavior I've noticed when upgrading from 0.34 to 1.0.

sinclairzx81:
@p2004a Hi, thanks for reporting.

I have published a fix for this on 1.0.80 that should resolve the issue.

> Is there some reason why the original options on Ref are not passed around?

Ah, this issue is mostly an oversight. As you saw, the `Instantiate` function attempts to extract generic type parameters for any given type and then proceeds to try and instantiate them, but in the Ref case, a non-resolvable Ref target should  resolve back to the original Ref + options (where the Ref may be instantiated later)

The correct behavior should be to retain the Ref options for non-resolvable targets, so this was fixed up on 1.0.80.

---

Will close out the issue for now as this should be resolved, but feel free to ping on this thread if you run into any problems.
Cheers!
S

---

BTW: Big fan of Beyond All Reason! It's an amazing RTS! :)

p2004a:
Thanks a lot for the quick fix! And happy you like the game 😄

PR #1524, opened 2026-01-26
Version 1.0.80

sinclairzx81:
This PR applies a fix to ensure `TRef` options are retained on calls to instantiate.


sinclairzx81:
Reference: https://github.com/sinclairzx81/typebox/issues/1522

Issue #1230, opened 2025-04-17
Additional properties flag ignored in Value.Parse()


omrishaked:
Setting `additionalProperties` to `false` works with `Value.Check()` but doesn't throw an error in `Value.Parse()`.  See example:
```
const schema = Type.Object(
  {
    a: Type.String(),
    b: Type.Number(),
  },
  {
    additionalProperties: false,
  },
);

console.log(`Should be true: ${Value.Check(schema, { a: 'Hello', b: 123 })}`);
console.log(`Should be false: ${Value.Check(schema, { a: 'Hello', b: 123, c: 'world' })}`);

console.log(`Should not throw: ${JSON.stringify(Value.Parse(schema, { a: 'Hello', b: 123 }))}`);
console.log(`Should throw: ${JSON.stringify(Value.Parse(schema, { a: 'Hello', b: 123, c: 'world' }))}`);
```

Output:
```
Should be true: true
Should be false: false
Should not throw: {"a":"Hello","b":123}
Should throw: {"a":"Hello","b":123} (doesn't throw)
```

I could use `Value.Check()` but then I wouldn't get the specific error that means my object has additional properties, I'd just get `false`.

PR #1532, opened 2026-02-20
Version 1.1.0

sinclairzx81:
# 1.1.0

## Overview

Version 1.1.0 introduces a behavioral change to `Value.Parse(...)` and `Validator.Parse(...)` where automatic corrective parsing is now disabled by default. This change addresses performance overhead when handling invalid data and prepares the `Value.*` API for future revisions.

This update introduces a validation semantics change and as such is published under a minor semver version.

## Corrective Parse

In 1.0, a feature referred to as “Corrective Parse” was trialed. This feature meant that for any invalid value passed to Parse, TypeBox would attempt to Convert, Default, and Clean the value, then re-Check it before throwing an assertion. 

The feature was introduced in 1.0 in an effort to establish a de facto parsing pipeline for TypeBox, but it has since been observed that it incurs significant performance overhead for large invalid values. Additionally, the correction behavior has generally proven surprising to many users, and is therefore better expressed as an opt-in feature.

The following shows the changes from 1.0 to 1.1

```typescript
// 1.0 
const A = Value.Parse(Type.Number(), '12345') // A = 12345

// 1.1
const A = Value.Parse(Type.Number(), '12345') // throw! - Expected Number
```

Corrective parsing can be re-enabled by using the `{ correctiveParse: true }` configuration.

```typescript
import { Settings } from 'typebox/system'

Settings.Set({ correctiveParse: true })

// 1.1
const A = Value.Parse(Type.Number(), '12345') // A = 12345 - OK
```

---

A new compiler parsing system for Value.* will be explored in subsequent revisions later in the year.

### DIFF
--- a/changelog/1.1.0.md
+++ b/changelog/1.1.0.md
@@ -1,0 +1,32 @@
+# 1.1.0
+
+## Overview
+
+Version 1.1.0 introduces a behavioral change to `Value.Parse(...)` and `Validator.Parse(...)` where automatic corrective parsing is now disabled by default. This change addresses performance overhead when handling invalid data and prepares the `Value.*` API for future revisions.
+
+This update introduces a validation semantics change and is published under a minor semver version.
+
+## Corrective Parse
+
+In 1.0, a feature referred to as “Corrective Parse” was trialed. This feature meant that for any invalid value passed to Parse, TypeBox would attempt to Convert, Default, and Clean the value, then re-Check it before throwing an assertion. The feature was introduced in 1.0 in an effort to establish a de facto parsing pipeline for TypeBox, but it has since been observed that it incurs significant performance overhead for large invalid values. Additionally, the correction behavior has generally proven surprising to users, and is therefore better expressed as an opt-in feature.
+
+The following shows the changes from 1.0 to 1.1
+
+```typescript
+// 1.0 
+const A = Value.Parse(Type.Number(), '123') // A = 123
+
+// 1.1
+const A = Value.Parse(Type.Number(), '123') // throw! - Expected Number
+```
+
+Corrective parsing can be re-enabled by using the `{ correctiveParse: true }` configuration.
+
+```typescript
+import { Settings } from 'typebox/system'
+
+// 1.1
+Settings.Set({ correctiveParse: true })
+
+const A = Value.Parse(Type.Number(), '123') // A = 123
+```
--- a/design/website/docs/system/1_settings.md
+++ b/design/website/docs/system/1_settings.md
@@ -48,11 +48,11 @@ export interface TSettings {
   useEval: boolean
 
   /**
-   * Enables or disables 'exactOptionalPropertyTypes' check semantics. By default, TypeScript 
-   * allows optional properties to be assigned 'undefined'. While this behavior differs from the 
-   * common interpretation of 'optional' as meaning 'key may be absent', TypeBox adopts the default 
-   * TypeScript semantics to remain consistent with the language. This option is provided to align 
-   * runtime check semantics with projects that configure 'exactOptionalPropertyTypes: true' in 
+   * Enables or disables 'exactOptionalPropertyTypes' check semantics. By default, TypeScript
+   * allows optional properties to be assigned 'undefined'. While this behavior differs from the
+   * common interpretation of 'optional' as meaning 'key may be absent', TypeBox adopts the default
+   * TypeScript semantics to remain consistent with the language. This option is provided to align
+   * runtime check semantics with projects that configure 'exactOptionalPropertyTypes: true' in
    * tsconfig.json.
    * @default false
    */
@@ -63,5 +63,15 @@ export interface TSettings {
    * @default false
    */
   enumerableKind: boolean
+  
+  /**
+   * Controls whether TypeBox uses corrective Parse. When enabled, TypeBox will attempt to recover invalid
+   * values during Parse by running a sequence of `Value.*` operations to `Convert`, `Default`, and `Clean`
+   * the value, followed by a subsequent `Assert`. Enabling this option may incur significant performance
+   * overhead for invalid values. It is recommended to keep this disabled in performance-sensitive
+   * applications.
+   * @default false
+   */
+  correctiveParse: boolean
 }
 ```
--- a/design/website/docs/value/parse.md
+++ b/design/website/docs/value/parse.md
@@ -1,8 +1,6 @@
 # Parse
 
-The Parse function attempts to parse a value and throws an error if the value is invalid. This function is similar to Decode, but it does not execute Decode callbacks. Parse is considered a faster version of Decode, as operations that transform values are skipped when the value already matches the expected type.
-
-> The Parse function first checks a value against the provided type and returns immediately if it matches. If the value does not match, it is processed through a sequence of Clone, Clean, Convert, and Default operations, and then re-checked. If the value remains invalid, a ParseError error is thrown.
+The Parse function validates and returns a value that conforms to a given type. If the value does not satisfy the type, a parse error is thrown.
 
 ## Example
 
@@ -11,5 +9,31 @@ Example usage is shown below.
 ```typescript
 const R = Value.Parse(Type.String(), 'hello')      // const R: string = "hello"
 
-const E = Value.Parse(Type.String(), [{ x: 1 }])   // throws ParseError 
+const E = Value.Parse(Type.String(), 12345)        // throws ParseError 
+```
+
+## Corrective Parse
+
+TypeBox provides an optional corrective parsing mode that attempts to repair invalid values before failing. When enabled, the parser runs a pipeline consisting of Convert, Default, and Clean, then re-asserts the value after processing. This feature can be useful when parsing environment variables into target types.
+
+> ⚠️ This feature can impact performance. It is not recommended for use in high throughput applications.
+
+This feature can be enabled as follows:
+
+```typescript
+import { Settings } from 'typebox/system'
+
+// Corrective Parse: Enable
+
+Settings.Set({ correctiveParse: true })
+
+// Corrective Parse: Convert Value into the target type if reasonable conversion is possible.
+
+const R = Value.Parse(Type.String(), 'hello')      // const R: string = "hello"
+
+const S = Value.Parse(Type.String(), 12345)        // const S: string = "12345"
+
+// Corrective Parse: Reset (optional)
+
+Settings.Reset()
 ```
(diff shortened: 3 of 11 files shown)


RobWelbourn:
I notice that the 1.1 version also removes the application of default values in the Parse pipeline -- I take it this is also the intended behavior?

In 1.x, what's the best way of creating a custom Parse pipeline?  I'm writing code that validates configuration files, and I'd like to be able to throw a ParseError for `{ additionalProperties: false }`.  The method described in [Additional properties flag ignored in Value.Parse() #1230
](https://github.com/sinclairzx81/typebox/issues/1230), doesn't seem to allow the insertion of a custom pipeline as the first parameter of `Value.Parse()`.  For what it's worth, `Value.Check()` behaves the same way -- it doesn't report on `additionalProperties`.

sinclairzx81:
@RobWelbourn Hi,

> I notice that the 1.1 version also removes the application of default values in the Parse pipeline -- I take it this is also the intended behavior?

Yes, this is intended. Moving forward, the functions Assert, Check and Parse will map to the following TS signatures.

```typescript
// Assert and Throw
function Assert<Type extends TSchema>(type: Type, value: unknown): asserts value is Static<Type>
// Check and Guard
function Check<Type extends TSchema>(type: Type, value: unknown): value is Static<Type>
// Check and Return
function Parse<Type extends TSchema>(type: Type, value: unknown): Static<Type>
```

... and where Parse is reduced to an simple operation that returns the value unmodified if valid.

---

> In 1.x, what's the best way of creating a custom Parse pipeline? 

The following will be the recommended way to construct a custom Parse pipelines moving forward.

[TypeScript Reference Link](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBANQIYBsCuBTANHA3nABWDAxWADsM4BfOAMyghDgHIYBPEgIwgA8B6AG6pMLAFChIsOABVOVBk1YdufcWIDGEcgGd4BJFB0YiJMpTgBeQsVIUMACgDaYuHH784AYSYgM5eCRyABM4KAwAWmhgjCg4JB04Sg0MHR1Ddlc4BwB9LQCMXhgcHJVsOGF0DABKKwA+RBEMADovFG1HSsxqrCyHfJhC4rgynC6a+saq5oARDDokNBQYfu1BopxRiqaevoGhzfkxncnkaZ9yQViV-Y2Ro+2q3bdVgrut8drLBrPMVpQMEFXuthh8dr03B44AAlNDkQiGYzxRJ0CioOAQEhQJAwYDaOB7NYHe4kY5PU5NZoGIyOW6gh6fMQAXWqYkKUngdDhGlx+OpxgAPHISHAhv5golhS0ZABlDQACwwICQdQcZQAXLIGU1NXCANbkCAAd3I1U1UuaMpgOOAGiF8gauCy4RgaCg8P5Jls5kcYPJCSSGCuUDE1DEYihESj0Zjsbj8YTiaTEc8AFV0gBzDApuBJvP5guJzTaPRhVJLeDWT0OC1WqAUDMOHpwACMACYAMwAFlZxd0EABzXajfCOgr1SAA)

```typescript
import Value, { Pipeline } from 'typebox/value'
import Type from 'typebox'

const ParsePipeline = Pipeline([
  // Comment and re-order as necessary
  (_context, _type, value) => Value.Clone(value),
  (context, type, value) => Value.Default(context, type, value),
  (context, type, value) => Value.Convert(context, type, value),
  (context, type, value) => Value.Clean(context, type, value),
  // Run Parse as final operation 
  (context, type, value) => Value.Parse(context, type, value)
])
export function Parse<Type extends Type.TSchema>(type: Type, value: unknown): Type.Static<Type> {
  return ParsePipeline(type, value) as never
}

// -------------------------------------------------------
// Usage
// -------------------------------------------------------
const result = Parse(Type.String(), 1234)

console.log(result)
```

Pipeline Documentation: https://sinclairzx81.github.io/typebox/#/docs/value/pipeline

---

Hopefully the Parse update hasn't caused too many issues. I had actually been planning to reduce to Parse operation to a simple check and return later this year (as I felt corrective parsing was a mistake), but the reference benchmark highlighting performance overhead prompted an earlier update. 

Technically the 1.1 Parse is going to be stricter (which is generally better), and the Pipeline API should hopefully prove a bit more 'opt-in' and make it more obvious what operations are being run, Getting this API right is been overly challenging, but I think the Parse operation in 1.1 is correct (as it fits into the Pipeline model)

Happy to field any additional questions on the update.
Cheers
S

RobWelbourn:
Thanks for your speedy response.  The suggested solution works great.  

Now, all I have to do is figure out how to identify the additional properties in `must not have additional properties`...

sinclairzx81:
@RobWelbourn Hi, all good.

> Now, all I have to do is figure out how to identify the additional properties in must not have additional properties...

If you need to catch on additional properties, you can omit the `Clean` and specify `additionalProperties: false` on the schema / type. This will throw if the incoming data has additional properties. 

```typescript
const ParsePipeline = Pipeline([
  (_context, _type, value) => Value.Clone(value),
  (context, type, value) => Value.Default(context, type, value),
  (context, type, value) => Value.Convert(context, type, value),
  //(context, type, value) => Value.Clean(context, type, value),
  (context, type, value) => Value.Parse(context, type, value)
])
export function Parse<Type extends Type.TSchema>(type: Type, value: unknown): Type.Static<Type> {
  return ParsePipeline(type, value) as never
}

// -------------------------------------------------------
// Usage
// -------------------------------------------------------
const result = Parse(Type.Object({ // throws!
  x: Type.Number(),
  y: Type.Number(),
}, { additionalProperties: false }), {
  x: 1,
  y: 2,
  z: 3
})
```
There would be ways to introspect a schema inside the custom Parse function and return different pipelines depending on the type / schema passed. Conditional / programmatic pipelining will be actually be explored later this year (along with Value.* compilation support), so if you have thoughts on features you would like to see, let me know!

sinclairzx81:
@RobWelbourn Hi, Just a head up.

Documentation on the `Value.Pipeline` has been published at the following URL

https://sinclairzx81.github.io/typebox/#/docs/value/pipeline

AshotN:
Does this mean TypeBox doesn't follow semvar? This was a breaking change.

sinclairzx81:
@AshotN Hi,

> Does this mean TypeBox doesn't follow SemVer? This was a breaking change.

Yes, you're quite right. I guess this should have been a major version bump under strict semantic versioning, and I have been thinking about this for much of today. Mostly, I have been considering whether SemVer can work for TypeBox while it attempts to stabilize Value.* (and other internals) in the 1.x line...

---

The update here implemented the `{ correctiveParse: true }` option to preserve previous behavior. Technically, no functionality was lost and backward compatibility was retained via configuration; and the configuration was provided with some mindfulness of being partially inline with SemVer minor backwards compat.

There are many updates planned for the 1.x stabilization work, I am not sure SemVer is ideal in the short to medium term, as majors lose all significance if they are triggered by relatively minor stabilization updates. But where I would like to roll out updates progressively (same for 0.x)

---

In all, I feel the consequence of SemVer would be that bug fix turnarounds and general release cycles would need to slow down in order for updates to queue up for a major/minor release. Note, there is a longer-term maintenance cost to that which I find difficult to balance as a single maintainer (I do prefer fixes to be fast and value the agility afforded by minor revision updates).

---

What are your thoughts on this? Should TypeBox opt for strict SemVer? I am interested in getting some thoughts and insights.

Apologies for the breakage
S

AshotN:
While I understand that something like this was needed, I was under the impression that 1.x meant stable release and that breaking changes would be reduced.

I think you should have kept `correctiveParse: true` by default and warn that a future version will flip that to `false`. 

But I understand it's not easy to manage all that alone, if you make it clear that semvar isn't guaranteed that's fine as well.

sinclairzx81:
@AshotN Hi, thanks for the follow up.

---

> While I understand that something like this was needed, I was under the impression that 1.x meant stable release and that breaking changes would be reduced.

Yeah, this was the high level plan but there's some additional details around the 1.x line. I will try and provide a bit more context.

The 1.x version was a broad re-architecture of the library, and backwards compat was just not going to be possible. The major revision was mostly to signal that there had been significant breaking between 0. x and 1.x (more than any other version of TypeBox), and with some mindfulness that 0.x was still widely used and no clear path to migrate users.

I had a bit of a logistical challenge.

- I need to support the 0.x line indefinitely (given the install base of [0.34.x](https://github.com/sinclairzx81/typebox-legacy))
- I need to advance the 1.x line (without breaking legacy users)

The decision to swap from `@sinclair/typebox` to `typebox` and distinguish between 0.x and 1.x is mostly done to make a clear distinction between two separate lineages of TypeBox, and to provide a slow and gradual opt-in path for users moving over. I had actually considered publishing 0.35.x to `typebox` (keeping the pre-1.0 iteration line), but as, the architecture between 0.x and 1.x was so vastly different, the 1.x seemed fitting.

---

> I think you should have kept correctiveParse: true by default and warn that a future version will flip that to false

Yeah, this would have been the expectation in hindsight. Apologies again. 

Historically, TypeBox tended to pushed low impact breaking changes like this on minor revisions, and the 1.1.x update was mostly following that release methodology, (albeit with a greater consideration to ensuring backwards compat options). I don't think this methodology is likely to work in a strictly SemVer sense, so still thinking about a iterative release strategy for the 1.x line.

---

> But I understand it's not easy to manage all that alone, if you make it clear that semvar isn't guaranteed that's fine as well.

Yeah, I think I will need to include something in the documentation around this. I will give it some thought before committing to anything. Getting TB on semver is the ideal, but given the nature of the library it will be difficult to manage (I think I would need to adopt queued up release cycles to do it properly)

All said, TypeBox 1.x is 99% stable (as an engineering metric). There's been a few small issues here and there post 1.x release (which was expected given the size of the update), but mostly you can expect good stability throughout. Just keep an eye out for `@deprecated` on 1.x line as there are a few things being removed in 2.x that were kept to reduce 0.x to 1.x friction, but these won't be officially removed until the next major.

Also, as a heads up, the main parts of the codebase that will be iterated on for 1.x relates to `Value.*`. The interface for `Value.*` should remain mostly the same, but the internals will be re-written against JSON Schema keyword semantics vs the current `Type.*` coupling (0.x). The only significant breaking changes being planned relates to Codecs, but will probably defer the break until 2.x. The Parse break in 1.1 is mostly in service for planned Codec changes later in the year.

Once 2.x is completed and Value.* moved over to 1.x's JSON Schema view of the world, that would effectively put TypeBox in a much better position for strict Semver adherence (and possibly even being able to split the library across multiple packages), but do need a bit of flexibility on the 1.x line to get there.

---

Thanks for the feedback, Feel free to ping if you have any questions on any of this (it help me understand the things I should be documenting and making more clear), all feedback is welcome and appreciated.

All the best!
S

PR #1534, opened 2026-02-25
fix(Schema): typo in `Schema.Validator.prototype.Parse`

mohamed-abdelnour:
Hello, @sinclairzx81! Thank you for all your work!

1. This fixes a typo in the Schema module in `Schema.Validator.prototype.Parse` that prevents error context from being reported correctly.

2. While browsing the code, to see why that typo was not caught by the type checker, I noticed a duplicate `export`, which I removed.

3. In case it is more convenient for you, I also bumped the patch version; let me know if I should drop that change from this pull request.

I made sure the following (Deno) tasks execute successfully (in my environment):

- `build`
- `format`
- `test`
- `turing`

### DIFF
--- a/tasks.ts
+++ b/tasks.ts
@@ -8,7 +8,7 @@ import { Range } from './task/range/index.ts'
 import { Metrics } from './task/metrics/index.ts'
 import { Task } from 'tasksmith'
 
-const Version = '1.1.1'
+const Version = '1.1.2'
 
 // ------------------------------------------------------------------
 // Build


sinclairzx81:
@mohamed-abdelnour Hi! thanks for the PR!

> This fixes a typo in the Schema module in Schema.Validator.prototype.Parse that prevents error context from being reported correctly.

Yeah, this is a good catch! Everything looks great here. Also thanks for finding the duplicate export. Will merge this one through and publish out shortly.

sinclairzx81:
@mohamed-abdelnour All good :)

Updates published out on 1.1.2. Thanks again for the PR!

All the best!
S

mohamed-abdelnour:
Thank you!

Issue #1444, opened 2025-11-05
Type.Partial mutates source schema and makes fields with Type.Base permanently optional


michealapach-del:
As soon as a schema create by `Type.Evaluate` get used with Type.Partial, it will turn the field that use Type.Base into `optional` forever:

```ts
export class TDateType extends Type.Base<Date> {

    public override Default(value: unknown): unknown {
        return new Date();
    }

    public override Create(): Date {
        return new Date();
    }

    public override Check(value: unknown): value is Date {
        return value instanceof Date
    }
    public override Errors(value: unknown): object[] {
        return !this.Check(value) ? [{ message: 'not a Date' }] : []
    }
}

export const BaseModel = Type.Object({
    status: Type.String(),
    updatedDate: DateType(),
    createdDate: DateType(),
});

const DataSchema = Type.Evaluate(
    Type.Intersect([
        BaseModel,
        Type.Object(
            {
                count: Type.Number()
            },
            { additionalProperties: false }
        )
    ])
);

const PartialSchema = Type.Object({
    freeSchema: Type.Partial(DataSchema) // it will make BaseModel optional forever
});

// from now, all schema create by BaseModel will loss required
const OtherDataSchema = Type.Evaluate(
    Type.Intersect([
        BaseModel,
        Type.Object(
            {
                count: Type.Number()
            },
            { additionalProperties: false }
        )
    ])
);

console.log(OtherDataSchema)
// {
//   type: 'object',
//   required: [ 'status', 'count' ],
//   properties: {
//     status: { type: 'string' },
//     updatedDate: TDateType { '~guard': [Object] },
//     createdDate: TDateType { '~guard': [Object] },
//     count: { type: 'number' }
//   }
// }
```

sinclairzx81:
@michealapach-del Hi, thanks for reporting. This does look like an oversight. 

The problem with `Type.Base` is that in order for the compositor to apply Optional (and remove Optional) modifiers, it needs to be able to `Clone` the Base. The problem is that `Type.Base` is a class instance, and there is no safe way to automatically `Clone` a class instance 

The only solution I can think of is to require implementations of Type.Base to implement a Clone override, as follows.

```typescript
export class TDateType extends Type.Base<Date> {
  public override Default(value: unknown): unknown {
      return new Date();
  }
  public override Create(): Date {
      return new Date();
  }
  public override Check(value: unknown): value is Date {
      return value instanceof Date
  }
  public override Errors(value: unknown): object[] {
[...]

PR #1582, opened 2026-04-18
Version 1.1.28

sinclairzx81:
This PR carries out broad optimization work on the TypeBox compositor. The update ensures computed types (Partial, Required, Pick, Omit, KeyOf, etc) by-pass the Instantiate(...) type, and initialize directly without exhaustive type matching. The PR also updates the internal CanInstantiate() type to perform a single depth check (vs deep Ref checking) and relies on the TRef instantiate to terminate at un-resolvable TRef.

> Other updates include deprecating ReadonlyType (this has been renamed to ReadonlyObject) and a minor regression for Base which will no longer Clone via the compositor. Users notified via https://github.com/sinclairzx81/typebox/issues/1444

These updates dramatically simplify internal Instantiate(...) logic, and make provisions for additional optimizations for Readonly / Optional, as well as Module / Cyclic.


SimenB:
Was this breaking change released as a patch on purpose? We're getting both type errors and runtime errors ("TypeError: Type.ReadonlyType is not a function") after upgrading

sinclairzx81:
@SimenB Hi, long time no hear, hope you're well :)

---

> Was this breaking change released as a patch on purpose? We're getting both type errors and runtime errors ("TypeError: Type.ReadonlyType is not a function") after upgrading

The `ReadonlyType` function name was flagged as `@deprecated`  on this patch [Ref](https://github.com/sinclairzx81/typebox/pull/1582/changes#diff-95b844fc1670750f5d18ab748901bbf3e1804ef1611e018822f71813ff48218dR57-R61), but the function was retained throughout 1.1.x and 1.2.x as an alias to `ReadonlyObject` (the function is still very much supported)

The `ReadonlyType` function was dropped on the 1.3.0 revision [Ref](https://github.com/sinclairzx81/typebox/pull/1636/changes#diff-ffcfb2bb995eff68da65d12aef73e22c1ac6fd04500223d8fcc5017d371b3755L56-L61) along with some other deprecations, but I don't recall explicitly removing it (as the other deprecations were getting the most attention for the 1.3.0 release), so looks to be an oversight.

---

### Rename Change

The rename from `ReadonlyType` to `ReadonlyObject` was to try and better capture the semantics of the function which are only meaningful to Object, Array and Tuple types. The original `Type` implied the operation was meaningful to non Object. I recall giving the name some consideration on this PR (as there was a lot of focus on trying to narrow operations to specific types), but didn't intend to remove the function, just alias the name.

---

The removal of the `ReadonlyType` function looks it was an oversight when going through other deprecations needed for 1.3.0. It is a undocumented change so would be happy to bring the function back as a bug fix on 1.3.5. 

Would you like me to revert the change?

sinclairzx81:
@SimenB Hiya,

Have gone ahead and reverted `ReadonlyType` on 1.3.5. This type shouldn't have been dropped on 1.3.0 (at least not without documentation) but got mixed up with the other deprecations related to Base, Promise and Iterators. The intention to was limit breakage on 1.3.0 to these types specifically.

> Deprecations on 1.3.0 were unfortunate but have been trying to stabilize the 1.x architecture and API surface to allow better semver adherence in 2.x. TB is still operating under the minor revisions for minor breaking changes under the 1.x revision line as that's proven the most practical given the scope of the library, but the Type.* API surface will be locked down for 2.x.

The ReadonlyType is still flagged as `@deprecated`. Will either drop the type in 1.4 or defer to 2.x which is being planned for later this year or early next.

Apologies for the breakage, but thanks for letting me know.
Cheers
S

SimenB:
Hey!

Yeah, I'm doing good, thanks 😀 

And thanks for the quick turnaround - 1.3.5 works correctly for us 👍

PR #1593, opened 2026-05-01
fix: prototype pollution in Value.Diff, Value.Patch, and Value.Clone

homanp:
## Description

When `Value.Diff` encounters a property key containing `/` (e.g. `__proto__/format`), it builds a multi-segment JSON Pointer path without RFC 6901 escaping. When that path is later applied via `Value.Patch`, `Pointer.Set` traverses into `Object.prototype` and writes attacker-controlled values there.

This becomes critical because `TypeCompiler.Compile` reads properties like `format` from the schema object's prototype chain, so polluting `Object.prototype.format` causes attacker-supplied strings to be interpolated into generated `new Function()` code, leading to arbitrary code execution.

The full chain works from a purely P2 position (attacker controls only runtime values, no schema control needed):

```
Attacker JSON: { "__proto__/format": "PAYLOAD" }
  → Value.Diff generates path /__proto__/format
  → Value.Patch writes PAYLOAD to Object.prototype.format  
  → TypeCompiler.Compile(Type.String()) picks up format from prototype
  → new Function() executes PAYLOAD
```

Separately, `Value.Clone` uses bracket assignment (`result[key] = value`) which triggers the `__proto__` setter instead of creating an own property, causing per-object prototype injection.

This PR fixes the independent issue that together case this chain.

### DIFF
--- a/src/schema/pointer/pointer.ts
+++ b/src/schema/pointer/pointer.ts
@@ -40,6 +40,12 @@ function AssertCanSet(value: unknown): asserts value is Record<string, unknown> 
   if(!Guard.IsObject(value)) throw Error('Cannot set value')
 }
 // ------------------------------------------------------------------
+// Security
+// ------------------------------------------------------------------
+function IsDangerousSegment(segment: string): boolean {
+  return segment === '__proto__' || segment === 'constructor' || segment === 'prototype'
+}
+// ------------------------------------------------------------------
 // Indices
 // ------------------------------------------------------------------
 function IsNumericIndex(index: string): boolean {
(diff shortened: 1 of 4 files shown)


sinclairzx81:
@homanp Hi!

Will close out via https://github.com/sinclairzx81/typebox/pull/1594 and include you as a co-contributor there.

Again, thanks for the PR, these kinds of reviews are very much appreciated!
Cheers
S

homanp:
@sinclairzx81 would you be open to joining this: https://www.superagent.sh/pact

happy to discuss. this way we could do continuous work finding vulnerabilities before attackers. open to a quick chat?

PR #1598, opened 2026-05-15
Fix StaticUnion regression for non-tuple Array<TSchema>

jackyzha0:
In typebox 0.34, UnionStatic used a mapped-index-signature pattern `{[K in keyof T]: Static<T[K]>}[number]` that resolved correctly for both tuple types and general `TSchema[]` arrays.

In 1.x, StaticUnion was rewritten using head/tail destructuring with a `Result = never` default. Because `TSchema[]` does not match `[Left, ...Right]` (the array may be empty), `Static<TUnion<TSchema[]>>` falls through to `never`. This regresses any code that builds a union from a variable typed as a plain array:

```typescript
const errors = [Type.Object(...), Type.Object(...)];
const schema = Type.Union(errors);
type T = Static<typeof schema>; // resolves to `never` instead of the union
```

Fix: add an `Array<infer Item extends TSchema>` fallback that infers the element type when the input is not a tuple. Tuples still go through the fast recursive path; the new branch only fires for general arrays.

Adds static tests covering:
  - inline tuple inference (existing behaviour)
  - generic TSchema[] (must not resolve to never)
  - narrowed Array<TLiteral<1|2|3>> (must resolve to 1|2|3)

### DIFF
--- a/src/type/types/union.ts
+++ b/src/type/types/union.ts
@@ -39,7 +39,9 @@ import { type TProperties } from './properties.ts'
 export type StaticUnion<Stack extends string[], Direction extends StaticDirection, Context extends TProperties, This extends TProperties, Types extends TSchema[], Result extends unknown = never> = (
   Types extends [infer Left extends TSchema, ...infer Right extends TSchema[]]
     ? StaticUnion<Stack, Direction, Context, This, Right, Result | StaticType<Stack, Direction, Context, This, Left>>
-    : Result
+    : Types extends Array<infer Item extends TSchema>
+      ? Result | StaticType<Stack, Direction, Context, This, Item>
+      : Result
 )
 // ------------------------------------------------------------------
 // Type
--- a/test/typebox/static/type/union.ts
+++ b/test/typebox/static/type/union.ts
@@ -1,12 +1,41 @@
-import { type Static, Type } from 'typebox'
+import { type Static, type TSchema, Type } from 'typebox'
 import { Assert } from 'test'
 
-const T = Type.Union([
-  Type.Literal(1),
-  Type.Literal(2),
-  Type.Literal(3)
-])
-type T = Static<typeof T>
+// inline tuple inference
+{
+  const T = Type.Union([
+    Type.Literal(1),
+    Type.Literal(2),
+    Type.Literal(3)
+  ])
+  type T = Static<typeof T>
 
-Assert.IsExtendsMutual<T, 1 | 2 | 3>(true)
-Assert.IsExtendsMutual<T, null>(false)
+  Assert.IsExtendsMutual<T, 1 | 2 | 3>(true)
+  Assert.IsExtendsMutual<T, null>(false)
+}
+// non-tuple TSchema[] should still resolve to a union (not never)
+{
+  const Schemas: TSchema[] = [
+    Type.Literal(1),
+    Type.Literal(2),
+    Type.Literal(3)
+  ]
+  const T = Type.Union(Schemas)
+  type T = Static<typeof T>
+
+  Assert.IsExtendsMutual<T, unknown>(true)
+  Assert.IsExtendsNever<T>(false)
+}
+// narrowed element type via Array<TLiteral<...>> should resolve to the literal union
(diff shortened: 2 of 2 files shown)


sinclairzx81:
@jackyzha0 Hi, sorry for the delay getting to this PR

> In typebox 0.34, UnionStatic used a mapped-index-signature pattern {[K in keyof T]: Static<T[K]>}[number] that resolved correctly for both tuple types and general TSchema[] arrays.

This is a general signature change in 1.x. The 0.34.x variadic type parameters may be revisited in future using TS `const-modifiers`, but the 1.x line needs to keep type parameter handling as simple as possible (at least for now) as work is being done on 1.x to optimize inference internals for the new [TScript\<...\>](https://sinclairzx81.github.io/typebox/#/docs/script/overview) feature (which calls to each type function)

To support exterior TS-tuple types (as per example), TB would need to opt for the following.

*Generic Const Modifiers*

https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#const-type-parameters

But would need to reviewed for instantiation overhead

---

For now, TB needs types to be either inlined into the type, or asserted as a fixed sized TS-tuple like that matches the type signature. This is applicable to the following variadic types.

[Reference](https://www.typescriptlang.org/play/?target=99&module=7#code/JYWwDg9gTgLgBAFQJ5gKZwGZQiOByGFVAIwgA88AoSgYwgDsBneBOAXjgG1k0A6AZRhRg9AOYAKAJQAaREV4A5AK4hiqKFIC6cAIaM4dJvAD0xuAB4AtJbgKI8HXLQAhcnEJpqpuAHUISgBsAEzh6VFQQmAhdMDAApHdomAALdAwIAICIAHcRUUpTArN1bCgALiKiuBs8KFQdIIZ4rgRBYTFZBGVVdU08OGB9entdRkZgUXodYgD0KLgwHSgdEFQYdTgIDHcifG42vM7utSg+2gZmOABBdidUXgBhC6ElGiiNBE75ADUIYCCpJJzkY4M5bjx7gAxJT0N7ABjiT53Xi-f6A4GXB7g+QASXo6ygjFQb0RQMMlwAIti+ABVejw+ikjHwACi1PuCCUcVQpKAA)

```typescript
import Type from 'typebox'

const T = [Type.String(), Type.Number()] as const // <-- Not a TypeBox type

// Would need to apply to the following
//
// error:
//
// - 'readonly [TString, TNumber]' is not assignable to parameter of type '[TString, TNumber]'
const A = Type.Constructor(T, Type.Void())
const B = Type.Function(T, Type.Void())
const C = Type.Intersect(T)
const D = Type.Union(T)
const E = Type.Tuple(T)
```

But can be fixed with

[Reference](https://www.typescriptlang.org/play/?target=99&module=7#code/JYWwDg9gTgLgBAFQJ5gKZwGZQiOByGFVAIwgA88AoSgYwgDsBneBOAXjgG1k0A6AZRhRg9AOYAKAJQAaREV4A5AK4hiqKFIC6cAIaMucAPSG4zYABtzcehHg7EAITiE0s4kruNG6mKgAmlHByfAiCwmKyQVHRMbFxUca6+vYAbjrmwH7ORHBgOlA6IKi+ULyBwai8CMqq6pSa1IkA8gDWtAzMcACC7BW8AMIdQko0MNDiCLI8lQBqEJlSku1M8E4c07wAYkr0o8AME1PycwuSS3QrcP29GwCS9CXeoxPnQ3AAIjfyAKr0+-QvZadACiXxCSjA5lQLyAA)

```typescript
import Type from 'typebox'

const T = [Type.String(), Type.Number()] as [ // still not a TB type, but asserted
  Type.TString,                               // as a valid type parameter.
  Type.TNumber
]

// Ok
const A = Type.Constructor(T, Type.Void())
const B = Type.Function(T, Type.Void())
const C = Type.Intersect(T)
const D = Type.Union(T)
const E = Type.Tuple(T)
```

---

Will close this one out for now, as I can't change the signatures at this stage. Happy to field any follow up questions on this thread (or via issue)

Cheers
S

PR #1602, opened 2026-05-27
feat: add Nullable type action

MrMcLaine:
Adds Type.Nullable(T) as a first-class action producing T | null, complementing the existing NonNullable action.


sinclairzx81:
@MrMcLaine Hi, thanks for the PR!

> Adds Type.Nullable(T) as a first-class action producing T | null, complementing the existing NonNullable action.

This PR is exceptionally well presented (well done!), unfortunately I can't accept this one as TypeBox limits support to syntactic constructs and built-in TS utility types only. The supported TS utility types can be found at the link below ....

https://www.typescriptlang.org/docs/handbook/utility-types.html

---

### Nullable

I do actually think `Nullable<T>` would be quite a good idea, but without TypeScript providing a built-in utility type, I can't really support it without introducing semantics TypeScript could introduce later. In terms of the semantics, the rules for `NonNullable<T>` are somewhat interesting as the action is used to Exclude both `null` and `undefined` from a type....for example

```typescript
/**
 * Exclude null and undefined from T <--- undefined considered nullable
 */
type NonNullable<T> = T & {};
```
The implies the natural inverse of `NonNullable<T>` would be `T | null | undefined`, so TypeBox would likely need to apply `undefined` for the action, but where `undefined` itself has no representation in JSON Schema. This could be a bit of a problem for developers modelling specifically for the spec.

Due to uncertainly around Nullable, the historical recommendation has been for users to implement it manually in user-land with the following....

```typescript
type Nullable<T> = T | null // explicit TS type.

function Nullable<T extends TSchema)(type: T) {
  return Type.Union([type, Type.Null()]) // ... or include Undefined if required
}
```

TypeBox will likely need to keep recommending user-land implementations of Nullable until such time as TS provides a nominated implementation of it, at which point, TypeBox would absolutely provide support for the type.

---

Hopefully this lends insight into why TB can't include it. Your PR is very well written though (I'm very impressed), and usually contributions like this get through, but for now, the call is to keep the Nullable type implemented in user land. 

Will close this one out, but happy to field any additional questions on this thread.
Thanks again,
S

PR #1612, opened 2026-06-09
fix: avoid calling property default factory twice in Value.Default for objects

chatman-media:
## Problem

`Value.Default` processes object schemas by iterating each property and resolving a default when the current value is `undefined`. The implementation called `FromType` **twice** for each property that needed a default:

1. Once to compute `propertyValue` (used only to check whether the result is an _unassignable undefined_)
2. Again to actually assign `value[key]`

```ts
// src/value/default/from_object.ts (before fix)
const propertyValue = FromType(context, type.properties[key], value[key])  // call 1
const isUnassignableUndefined = ...
if (isUnassignableUndefined) continue
value[key] = FromType(context, type.properties[key], value[key])           // call 2 — bug
```

For **static** defaults (`default: 42`) this double-call was harmless. For **function-based** defaults (`default: () => …`) the factory ran twice per absent property: the first return value was silently discarded and the second was assigned.

This is a real bug for any factory that produces unique values on each call — auto-increment IDs, timestamps, UUIDs, or any stateful generator:

```ts
let counter = 0
const T = Type.Object({
  id: Type.String({ default: () => `id-${++counter}` }),
})

Value.Default(T, {})
// Expected: { id: 'id-1' }, counter === 1
// Actual:   { id: 'id-2' }, counter === 2   ← factory called twice
```

## Fix

Reuse the `propertyValue` already computed on the first call instead of invoking `FromType` a second time:

```ts
// src/value/default/from_object.ts (after fix)
const propertyValue = FromType(context, type.properties[key], value[key])
const isUnassignableUndefined = ...
if (isUnassignableUndefined) continue
value[key] = propertyValue   // ← reuse, no second call
```

## Tests

Two new test cases added to `test/typebox/runtime/value/default/object.ts`:

- **Should Default 33** — single factory property, verifies it is called exactly once and the correct value is assigned
- **Should Default 34** — two factory properties, verifies each is called exactly once

Both tests **fail** before the fix and **pass** after. All 34,879 existing tests continue to pass.

### DIFF
--- a/src/value/default/from_object.ts
+++ b/src/value/default/from_object.ts
@@ -49,7 +49,7 @@ export function FromObject(context: TProperties, type: TObject, value: unknown):
     if (isUnassignableUndefined) continue
 
     // Assign
-    value[key] = FromType(context, type.properties[key], value[key])
+    value[key] = propertyValue
   }
   // return if not additional properties
   if (!IsAdditionalProperties(type) || Guard.IsBoolean(type.additionalProperties)) return value
--- a/test/typebox/runtime/value/default/object.ts
+++ b/test/typebox/runtime/value/default/object.ts
@@ -354,3 +354,40 @@ Test('Should Default 32', () => {
   const R = Value.Default(T, {})
   Assert.IsEqual(R, { id: undefined }) // pick first union match
 })
+// ------------------------------------------------------------------
+// Factory Default: Should be called exactly once per absent property
+// ------------------------------------------------------------------
+Test('Should Default 33', () => {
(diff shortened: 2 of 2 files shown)


sinclairzx81:
@chatman-media Hi, thanks for the PR!

This was a excellent find! The update is both a fix and a optimization!! :) Nice work!

Will merge through and publish on `1.2.7` shortly

sinclairzx81:
@chatman-media Updates have been published out to `typebox@1.2.7`

Thanks again :) All the best!
S

Issue #1620, opened 2026-06-15
Value.Decode/Encode/Clean mutate Union's member


SaltyAom:
Calling `Value.Decode`, `Value.Encode`, and `Value.Clean` mutate schema's `anyOf`

This is not a documented behavior, and I'm not sure if this is intended behavior (from my understanding: sort is intended but mutate is probably not)

## 1. Value.Create

```js
import { Type } from 'typebox'
import { Value } from 'typebox/value'

const T = Type.Union([Type.Number(), Type.Literal(1)])

console.log(Value.Create(T)) // 0
Value.Clean(T, 0)
console.log(Value.Create(T)) // 1
```

`Value.Clean` mutate reorders `T.anyOf` from `[number, const:1]` to `[const:1, number]`, so `Value.Create` (which builds from `anyOf[0]`) is not consistent

Effect:
1. In Elysia, we are using `Value.Create` to generate `expect` value for error screen which make `default` not reliable
2. Swagger UI/Scalar OpenAPI documentation will have a problem with showing the incorrect default data because of the reordering

## 2 Non-deterministic JSON Schema

So basically, this will have a problem with identity for cache storage

```js
import { Type } from 'typebox'
import { Value } from 'typebox/value'

const make = () => Type.Object({ id: Type.Union([Type.Number(), Type.Literal(1)]) })
const A = make()
const B = make()

console.log(JSON.stringify(A) === JSON.stringify(B)) // true

Value.Decode(B, { id: 1 })

console.log(JSON.stringify(A) === JSON.stringify(B)) // false (because of reorder)
```

Effect:
If a user uses JSON schema as a key to cache a computation or has an equality check, it will basically break or skip the cache entirely
 
## Root cause and fix
This happens because `UnionPrioritySort` (in [src/value/shared/union_priority_sort.ts](https://github.com/sinclairzx81/typebox/blob/main/src/value/shared/union_priority_sort.ts#L59)) uses the `.sort()` method 

This should be able to be fixed by changing `sort` to [toSorted](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted) or `[...types].sort()` if you don't want to use ES2023 feature

sinclairzx81:
@SaltyAom Hi! Thanks for reporting.

> This is not a documented behavior, and I'm not sure if this is intended behavior (from my understanding: sort is intended but mutate is probably not)

[...]

PR #1626, opened 2026-06-16
perf(value): memoize UnionPrioritySort to avoid per-value JSON.stringify

### LINKED ISSUE #1624: Value.Clean is very slow on union-heavy schemas — UnionPrioritySort re-sorts per value
## Summary

`clean/from_union` calls `UnionPrioritySort(type.anyOf)` **on every value**, and
`UnionPrioritySort`'s comparator serializes each member *schema* with `JSON.stringify`
(`DeterministicCompare`). The sort order depends only on the (static) schema, never on the
value — so this is pure per-value waste, and it scales with the **serialized size** of the
union members. On schemas saturated with `Type.Union([X, Null])` (every nullable field) it
dominates `Value.Clean`, and therefore `Value.Decode`/`Encode` (which run `Clean`).

Note `DeterministicCompare` is not a rare tie-breaker: it runs on the `disjoint` branch of the
comparator, and `Compare(X, Null)` is `disjoint` for *every* nullable pair (`Object`/`String`/
`Enum` vs `Null`) — as is `Compare(Object, Object)`. `disjoint` ("neither member contains the
other") is the **normal** case for nullable and discriminated unions, so `JSON.stringify` fires
on essentially every union comparison. Measured: one `Value.Clean` over 1,000 records with ~6
nullable-union fields each made **~12,000 `JSON.stringify` calls** (~12/record).

## Regression (1.0.x → 1.2.x)

`value/clean/from_union`:

```js
// 1.0.41 — plain declaration order, no serialization
for (const schema of type.anyOf) {
  const clean = FromType(context, schema, Clone(value));
  if (Check(context, schema, clean)) return clean;
}

// 1.2.10 — sorts per value; comparator JSON.stringify's member schemas
for (const schema of UnionPrioritySort(type.anyOf)) { ... }
```

`value/shared/union_priority_sort.mjs`:
```js
function DeterministicCompare(left, right) {
  return JSON.stringify(left).localeCompare(JSON.stringify(right)); // serializes SCHEMAS
}
export function UnionPrioritySort(types, order = 1) {
  return types.sort((l, r) => { /* Compare(l,r); falls back to DeterministicCompare */ });
}
```
`codec/from_union` and `clean/from_union` both call it per value.

## Measurements (`typebox@1.2.10`, Node 22.22.3)

A schema whose fields are mostly `Type.Union([X, Null])` (every nullable field), several
wrapping sizable objects, N=10,000, `Value.Clean`: **1.0.41 ≈ 463 ms → 1.2.10 ≈ 6,054 ms
(~13×)** — `Clean` goes from a minor stage to the dominant one in `Value.Decode`.

Smoking gun — `Value.Clean` on a value that is **always `null`** (nothing to clean), so the
only size-dependent work is the sort's `JSON.stringify`, N=50,000:

| `Union([Object(w fields), Null])` | `Value.Clean` |
|---|---|
| w = 2 | ~124 ms |
| w = 10 | ~150 ms |
| w = 40 | ~256 ms |
| w = 100 | ~483 ms |

Cost grows with member-schema *width* on a constant null value → it's the serialization, not
any cleaning.

## Impact and reach

`UnionPrioritySort` is called by exactly two modules — `value/clean/from_union.mjs` and
`value/codec/from_union.mjs` — so memoizing it speeds up every pipeline that cleans or
codec-walks a union: **`Value.Clean` directly, and therefore `Value.Parse`, `Value.Decode`,
`Value.Encode` (all run `Clean`)**, plus the codec walk's branch selection. (`Default`, `Convert`
and `Check` use different union handling and are unaffected.)

Measured by memoizing `UnionPrioritySort` by the `anyOf` array identity (same output) and
re-running the same codec schema (an array of objects with many `Union([X, Null])` fields + two
codec leaves), count=10,000, median of 7:

| | today | memoized |
|---|---|---|
| `Value.Clean` | ~1,543 ms | ~291 ms |
| codec walk (`Value.DecodeUnsafe`) | ~1,535 ms | ~283 ms |
| `compiled.Decode` (full interpreted path) | ~4,008 ms | ~1,424 ms (**2.8×**) |

Combined with the companion compiled-codec-path issue (#1625), codec `Decode` on that schema drops
to ~302 ms (~13×); see that issue's impact table.

## Fix

The priority order is a pure function of the static schema. Compute it once and cache it
(e.g. memoize `UnionPrioritySort` by the `anyOf` array identity, or precompute sorted members
when the type is built/compiled), instead of sorting + `JSON.stringify`-ing per value.
Same output, removes the per-value serialization. Affects `clean` and `codec` `from_union`.
(Verified: memoizing by `anyOf` identity drops `Value.Clean` ~1,543 → ~291 ms on the schema
above, output unchanged.)

## Repro
`clean-regression.mjs` (the smoking gun above) — standalone, only `typebox`, Node ≥ 22.

## Reproduction scripts (standalone, only `typebox`, Node ≥ 22)

<details>
<summary><code>clean-regression.mjs</code></summary>

```js
/**
 * Smoking gun for the `clean/from_union` regression.
 *
 * `Value.Clean`'s cost on a union scales with the SERIALIZED SIZE of the union's
 * member schemas, even when the value is `null` (so there is nothing to clean) —
 * because `UnionPrioritySort` calls `JSON.stringify` on each member schema, on
 * every value, via its `DeterministicCompare` tie-break (which runs on the
 * `disjoint` branch, i.e. the normal case for nullable unions). The sort order
 * depends only on the static schema, so doing it per value is pure waste.
 *
 * Run on Node >= 22:  node clean-regression.mjs [count]
 */
import Type from 'typebox';
import * as Value from 'typebox/value';
import { performance } from 'node:perf_hooks';
import { readFileSync } from 'node:fs';

const COUNT = Number(process.argv[2] ?? 50_000);
const typeboxVersion = JSON.parse(
  readFileSync(new URL('./node_modules/typebox/package.json', import.meta.url))
).version;

// A wide object schema with `fieldCount` string properties.
const wideObject = (fieldCount) =>
  Type.Object(Object.fromEntries(Array.from({ length: fieldCount }, (_, i) => [`f${i}`, Type.String()])));

// `{ items: Array(Union([wideObject, Null])) }`. The value is always null, so
// Clean has nothing to strip — the only size-dependent work left is the sort.
const schemaWithMemberWidth = (fieldCount) =>
  Type.Object({ items: Type.Array(Type.Union([wideObject(fieldCount), Type.Null()])) });

const allNullPayload = { items: Array.from({ length: COUNT }, () => null) };

const median = (fn, repetitions = 5) => {
  fn(); // warm
  const samples = [];
  for (let i = 0; i < repetitions; i++) {
    const start = performance.now();
    fn();
    samples.push(performance.now() - start);
  }
  samples.sort((a, b) => a - b);
  return samples[Math.floor(samples.length / 2)];
};

console.log(`typebox ${typeboxVersion} | count=${COUNT.toLocaleString()} | value is always null (Clean has nothing to strip)\n`);
console.log('  member-schema width        Value.Clean');
for (const fieldCount of [2, 10, 40, 100]) {
  const schema = schemaWithMemberWidth(fieldCount);
  const ms = median(() => Value.Clean({}, schema, Value.Clone(allNullPayload)));
  console.log(`  ${String(fieldCount).padStart(3)} fields                 ${ms.toFixed(1).padStart(8)} ms`);
}

console.log('\n  Clean time grows with member-schema SIZE on a constant null value:');
console.log('  the cost is JSON.stringify(memberSchema) inside UnionPrioritySort, per value.');
console.log('  The sort depends only on the static schema, so it should be computed once.');
```

</details>

amit-evron-cyera:
Fixes #1624.

## Problem

`UnionPrioritySort` sorts a union's members on **every call**, and its comparator
(`DeterministicCompare`) calls `JSON.stringify` on each member *schema* whenever `Compare`
returns `disjoint`. `disjoint` ("neither member contains the other") is the normal case for
`Type.Union([X, Null])` and discriminated unions, so the `JSON.stringify` runs on essentially
every comparison.

The sort order is a pure function of the (static) member schemas, but it is recomputed per
value. During `Codec` `Decode`/`Encode` and `Value.Clean` over large arrays of union-bearing
values, this repeats identical serialization work on every element. It dominates `Value.Clean`
(and therefore `Value.Decode`/`Encode`/`Parse`, which run `Clean`) on union-heavy schemas.

A minimal demonstration: `Value.Clean` on a payload that is always `null` (so there is nothing
to clean) still gets slower as the union member schema grows wider — the only size-dependent
work left is the comparator's `JSON.stringify`.

## Fix

Memoize the sorted result per source array and `order` (a `WeakMap` keyed by the `anyOf`
array). A fresh array is returned on each call, so:

- the source array is never mutated (preserves the #1620 fix from 1.2.12), and
- callers never observe a shared array (mutating a returned array can't corrupt the memo).

The comparator (and its `JSON.stringify`) now runs once per union schema instead of once per
value. Output is unchanged — same order, same fresh-array contract.

> Note: this builds on the 1.2.12 non-mutation fix (#1620). That change stopped the mutation
> but kept the per-value sort/serialization; this addresses the remaining cost.

An alternative shape would be to precompute the order when the type is built/compiled; the
`WeakMap` memo was chosen as the smaller, localized change. Happy to switch if you prefer.

## Performance (typebox built from this branch's `src`, Node 22, count = 10,000)

On a schema that is an array of objects with many `Union([X, Null])` fields + two codec leaves:

| | before | after |
|---|---|---|
| `Value.Clean` | ~1,543 ms | ~291 ms |
| codec walk (`Value.DecodeUnsafe`) | ~1,535 ms | ~283 ms |
| `compiled.Decode` (full interpreted path) | ~4,008 ms | ~1,424 ms |

## Tests

- Full runtime suite passes: `4417 passed | 0 failed`.
- Existing `UnionPrioritySort` tests (incl. the #1620 non-mutation test) unchanged and passing.
- Added two tests: result is stable across calls and returned as a distinct array per call
  (mutating a returned array does not affect later calls).

## Assumption

Caching by `anyOf` identity assumes a union's member array is treated as immutable after
construction — consistent with the direction in #1620. If a schema's `anyOf` is mutated in
place after first use, the memo would be stale; that is already an unsupported pattern.

### DIFF
--- a/src/value/shared/union_priority_sort.ts
+++ b/src/value/shared/union_priority_sort.ts
@@ -45,6 +45,20 @@ function DeterministicCompare(left: TSchema, right: TSchema): number {
   return JSON.stringify(left).localeCompare(JSON.stringify(right))
 }
 // ------------------------------------------------------------------
+// Sort
+// ------------------------------------------------------------------
+function Sort(types: TSchema[], order: number): TSchema[] {
+  return [...types].sort((left, right) => {
+    const result = Compare(left, right) as string
+    return (
+      Guard.IsEqual(result, 'disjoint') ? DeterministicCompare(left, right) :
+      Guard.IsEqual(result, 'right-inside') ? 1 :
+      Guard.IsEqual(result, 'left-inside') ? -1 :
+      DeterministicCompare(left, right)
+    ) * order
+  })
+}
+// ------------------------------------------------------------------
 // UnionPrioritySort
 //
 // Performs a deterministic sort on Union members. By default, this
@@ -53,17 +67,28 @@ function DeterministicCompare(left: TSchema, right: TSchema): number {
 // the order property to -1 which will reverse unions from broader
 // to more narrow.
 //
+// The sort order is a pure function of the (static) member schemas, so the
+// result is memoized per source array and order. This avoids re-running the
+// DeterministicCompare (JSON.stringify) comparator on every call, which is
+// significant when Codec Decode/Encode or Value.Clean process large arrays of
+// union-bearing values (the comparator otherwise serializes member schemas on
+// every value). A fresh array is returned on each call, so the source array is
+// never mutated (see #1620) and callers never observe a shared array.
+//
 // ------------------------------------------------------------------
+const memo = new WeakMap<TSchema[], Map<number, TSchema[]>>()
 
 /** Deterministically sorts schemas by structural relationship (narrow to broad) */
 export function UnionPrioritySort(types: TSchema[], order: number = 1): TSchema[] {
-  return [...types].sort((left, right) => {
-    const result = Compare(left, right) as string
-    return (
-      Guard.IsEqual(result, 'disjoint') ? DeterministicCompare(left, right) : 
-      Guard.IsEqual(result, 'right-inside') ? 1 : 
-      Guard.IsEqual(result, 'left-inside') ? -1 : 
-      DeterministicCompare(left, right)
-    ) * order
-  })
+  let byOrder = memo.get(types)
+  if (Guard.IsUndefined(byOrder)) {
+    byOrder = new Map<number, TSchema[]>()
+    memo.set(types, byOrder)
+  }
+  let sorted = byOrder.get(order)
+  if (Guard.IsUndefined(sorted)) {
+    sorted = Sort(types, order)
+    byOrder.set(order, sorted)
+  }
+  return [...sorted]
 }
--- a/test/typebox/runtime/value/shared/union_priority_sort.ts
+++ b/test/typebox/runtime/value/shared/union_priority_sort.ts
@@ -197,3 +197,20 @@ Test('Should UnionPrioritySort 26', () => {
   UnionPrioritySort(A) // Priority Sort (A)
   Assert.IsEqual(A, B) // Expect Equal
 })
+// ------------------------------------------------------------------
+// Memoized result is stable and returned as a fresh array per call
+// ------------------------------------------------------------------
+Test('Should UnionPrioritySort 27', () => {
+  const A = [Type.Number(), Type.Literal(1)]
+  const R1 = UnionPrioritySort(A)
+  const R2 = UnionPrioritySort(A)
+  Assert.IsEqual(R1, R2) // stable order across calls
+  Assert.IsTrue(R1 !== R2) // distinct array per call (callers never share)
+})
+Test('Should UnionPrioritySort 28', () => {
+  const A = [Type.Number(), Type.Literal(1)]
+  const R1 = UnionPrioritySort(A)
+  R1.push(Type.String()) // mutating a returned array must not affect later calls
+  const R2 = UnionPrioritySort(A)
+  Assert.IsEqual(R2, [Type.Literal(1), Type.Number()])
+})


sinclairzx81:
@amit-evron-cyera Hi, thanks for the PR

Unfortunately, I can't really add WeakMap memorization/caching internal to TypeBox on Value.* ops as they need to be as stateless as possible. The UnionPrioritySort is actually meant as infrastructure for a future Value.* compiler, so don't want to be adding side effects that need to be pulled out later (especially if it relates to performance)

For now, I have added a solution on https://github.com/sinclairzx81/typebox/pull/1627 which re-implements UnionPrioritySort as an exported standalone function. I have also added a configuration to disable implicit / automatic UnionPrioritySort. 

The update allows for memo/caching external to TypeBox.

```typescript
import Value, { UnionPrioritySort } from 'typebox/value'
import System from 'typebox/system'
import Type from 'typebox'

// Unsorted Union
const A = Type.Script(`
  { x: number } | 
  { x: number, y: number } |
  { x: number, y: number, z: number } |
`)

// *NEW* Manually Sort Union 
const B = UnionPrioritySort(A) // <--- CACHE (B) EXTERNAL TO TYPEBOX

// *NEW* Disable Automatic Priority Sort
System.Settings.Set({ unionPrioritySort: false })

// Benchmark
console.time()
for(let i = 0; i < 100_000; i++) {
  Value.Clean(B, { x: 1, y: 1, w: 3 })
}
console.timeEnd() // 737ms - IMPROVED

// Reset! 
System.Settings.Reset()
```

For TypeBox 1.2, I need to keep on with the implicit UnionPrioritySort for now, but will consider disabling it on the next revision (1.3) as the performance overhead is far too high (the functionality really should be opt-in)

---

Will close this one up for now, but thanks for taking a look.
Cheers
S

amit-evron-cyera:
Thank you for the quick response! I'm testing the this new control with my banchmark

PR #1627, opened 2026-06-17
Version 1.2.15

### LINKED ISSUE #1624: Value.Clean is very slow on union-heavy schemas — UnionPrioritySort re-sorts per value

sinclairzx81:
This PR implements an optimization for Clean, Decode and Encode by making the Value.* UnionPrioritySort logic configurable (opt out) as well as providing a public API for callers to preprocess and cache a type manually.

> Note: Given the performance overhead of implicit union sorting in Value.* paths, TypeBox will likely disable automatic `unionPrioritySort` in a future mainline revision, but needs to retain the implicit sort in 1.2.x to avoid a breaking change. This PR enables a fast path workaround.

Fixes: https://github.com/sinclairzx81/typebox/issues/1624

## Default

By default TypeBox automatically pre-processes types on each call to Clean, Encode and Decode. This pre-processing step is expensive but required for correct union matching. The following is the default performance for 100K iterations of Clean:

```typescript
// Unsorted Union Type
const A = Type.Script(`
  { x: number } | 
  { x: number, y: number } |
  { x: number, y: number, z: number } |
`)

// Benchmark
console.time()
for(let i = 0; i < 100_000; i++) {
  Value.Clean(A, { x: 1, y: 1, w: 3 })
}
console.timeEnd() // 4488ms - TOO SLOW
```

## Optimization

This PR enables callers to manually (and optionally) cache the sorted / pre-processed type next to compiled versions of the type. The PR adds a `unionPrioritySort` configuration option to System Settings which can be disabled, and the `UnionPrioritySort` function is exported as a member of Value.* to enable manual ordering outside of TypeBox.

```typescript
import Value, { UnionPrioritySort } from 'typebox/value'
import System from 'typebox/system'
import Type from 'typebox'

// Unsorted Union
const A = Type.Script(`
  { x: number } | 
  { x: number, y: number } |
  { x: number, y: number, z: number } |
`)

// *NEW* Manually Sort Union (This type can be cached)
const B = UnionPrioritySort(A)

// *NEW* Disable Automatic Priority Sort
System.Settings.Set({ unionPrioritySort: false })

// Benchmark
console.time()
for(let i = 0; i < 100_000; i++) {
  Value.Clean(B, { x: 1, y: 1, w: 3 })
}
console.timeEnd() // 737ms - IMPROVED

// Reset! 
System.Settings.Reset()
```

---

### Type Level Engine Priority 

This PR also adds the Priority sorting algorithm to the Type.* engine infrastructure, implemented at both runtime and statically. Automatic sorting of Union variants may be considered in future versions of TypeBox after the TypeScript 7 release.

The Priority API is available via selective import:

```typescript
import Type, { Priority } from 'typebox'

const A = Priority([
  Type.Script(`{ x: 1 }`),
  Type.Script(`{ x: 1, y: 1, z: 1 }`),
  Type.Script(`{ x: 1, y: 1 }`),
])

// const A: [Type.TObject<{
//     x: Type.TLiteral<1>;
//     y: Type.TLiteral<1>;
//     z: Type.TLiteral<1>;
// }>, Type.TObject<{
//     x: Type.TLiteral<1>;
//     y: Type.TLiteral<1>;
// }>, Type.TObject<{
//     x: Type.TLiteral<1>;
// }>]
```


sinclairzx81:
Ref: https://github.com/sinclairzx81/typebox/issues/1620

Issue #1625, opened 2026-06-16
Codec schemas make compiled Decode/Encode orders of magnitude slower than the codec work needs


amit-evron-cyera:
## Problem

When a schema contains a `Type.Codec`, `Compile(schema).Decode/Encode` delegates to the
fully-interpreted `Value.Decode/Encode` (the `hasCodec` branch in `Validator`) and never uses
the compiled validator it just built. Three costs:

1. **It asserts with the interpreted `Check`, not the compiled one.** This is structural, not an
   oversight: `Validator.Decode(value)` delegates to the module-level `Decode(context, type, value)`
   in `value/codec`, a standalone schema-only pipeline that never receives the compiled validator —
   so its `Assert` can only call the interpreted `Check`; the JIT `Check` lives on the `Validator`.
   (The two are equivalent: across a fuzz of valid/invalid inputs `compiled.Check` and the
   interpreted `Value.Check` agree 11/11, so using the compiled one is behaviour-preserving — it
   just requires the `Validator` to own a codec pipeline, which is what this issue proposes.)
2. **The codec walk visits every node, not just codec-bearing ones.** `DecodeUnsafe`→`FromType`
   type-dispatches and recurses through entirely codec-free subtrees of the *value*. Whether a
   subtree can contain a codec is a property of the *schema*, which is fixed — but it's never
   precomputed, so the walk re-discovers "nothing here" on every value. (`HasCodec`,
   `value/codec/has.mjs`, answers this per subtree but is only ever called once at the root —
   1.0.7 / #1331.)
3. **The codec walk re-discriminates unions per value.** `codec/from_union` runs an interpreted
   `Check` against each `anyOf` member to find the matching branch — the exact work the compiled
   `Check` already did:
   ```js
   // value/codec/from_union.mjs (Decode)
   for (const schema of UnionPrioritySort(type.anyOf, 1)) {
     if (!Check(context, schema, value)) continue;   // interpreted check, per member, per value
     return Callback(direction, context, type, FromType(direction, context, schema, value));
   }
   ```

The result is `O(value × whole-tree × union-arity)` interpreted work per call, even though the
only real work is the codec callbacks. (`Decode`'s interpreted pipeline also runs
`Clone/Default/Convert/Clean` unconditionally; a Check-first path skips those for valid input —
see Proposal.)

## Evidence (`typebox@1.2.10`, Node 22, standalone — only `typebox`)

[...]

PR #1631, opened 2026-06-18
perf(value/codec): prune the codec walk to codec-bearing paths via a StripToCodec preprocessor

amit-evron-cyera:
Follow-up to #1625, taking up the localized per-stage optimization you said you're open to.

**Why this is in scope now (not pre-empting the 2.x pipeline work):**
- You're making `DecodeUnsafe`/`EncodeUnsafe` the **primary** codec API in 2.x — so optimizing that specific step is directly on that path, and 2.x will want it regardless.
- It's a **localized optimization of one stage** (the codec walk), which you said you'd consider — same spirit as a faster `Convert` helping any pipeline.
- It **can't be done in user space** (below), so it's not something users can opt into themselves.
- It's **self-contained and stateless**, and can be **gated behind a `Settings` flag** (mirroring `unionPrioritySort`) if you'd prefer it opt-in.

### Idea
`DecodeUnsafe`/`EncodeUnsafe` walk the whole value, including subtrees that statically contain no codec. "Contains a codec" is a property of the *schema*, so resolve it once per call: `StripToCodec(type)` returns the schema reduced to its codec-bearing paths — Object/Array/Record children with no codec are dropped; Union/Intersect/Tuple/Ref/Cyclic are kept intact. The **existing walk runs unchanged** on the stripped schema and touches only codec paths.

- Once-per-call type preprocessor — same pattern and placement as `UnionPrioritySort`.
- No retained state, no caching; the walkers are untouched.
- The input schema is not mutated (verified by deep-freezing the schema and round-tripping) — preserves the #1620 contract.

### Why it can't be user-side
Pruning in user space means re-implementing the per-type codec traversal (`from_object`/`from_array`/`from_union`/…), and the 2.x Pipeline API doesn't close the gap — composing stages still calls the full-tree walk. So for codec users it's binary: either `DecodeUnsafe` prunes, or they pay the full-tree walk.

### Numbers
Schema with ~2 codec leaves in a large object (many `Union([X, Null])` fields + nested), decoded over a 10k-element array (Node 22): `Check + DecodeUnsafe` ~260 ms → ~8 ms, byte-identical.

### Correctness
- Full runtime suite passes (4434).
- Byte-identical to `Value.Decode` across the existing codec suite + added cases (extra keys, optional-with-default, wrong-type, nested object/array/union/tuple of codecs).
- Schema-immutability verified via deep-freeze round-trip.

If you'd like it opt-in, I'll add a `Settings` flag mirroring `unionPrioritySort`.

### DIFF
--- a/src/value/codec/decode.ts
+++ b/src/value/codec/decode.ts
@@ -43,6 +43,7 @@ import { Pipeline } from '../pipeline/index.ts'
 import { FromType } from './from_type.ts'
 
 import { UnionPrioritySort } from '../shared/union_priority_sort.ts'
+import { StripToCodec } from './strip.ts'
 
 // ------------------------------------------------------------------
 // Assert
@@ -61,7 +62,11 @@ function Assert(context: TProperties, type: TSchema, value: unknown): unknown {
 // ------------------------------------------------------------------
 /** Executes Decode callbacks only */
 export function DecodeUnsafe(context: TProperties, type: TSchema, value: unknown): unknown {
-  const sorted = Settings.Get().unionPrioritySort ? UnionPrioritySort(type) : type
+  // Strip to codec-bearing paths first, then sort only what survives — the sort
+  // (UnionPrioritySort) rebuilds the whole schema, so running it on the pruned
+  // schema is cheaper.
+  const stripped = StripToCodec(context, type)
+  const sorted = Settings.Get().unionPrioritySort ? UnionPrioritySort(stripped) : stripped
   return FromType('Decode', context, sorted, value)
 }
 // ------------------------------------------------------------------
--- a/src/value/codec/encode.ts
+++ b/src/value/codec/encode.ts
@@ -43,6 +43,7 @@ import { Pipeline } from '../pipeline/index.ts'
 import { FromType } from './from_type.ts'
 
 import { UnionPrioritySort } from '../shared/union_priority_sort.ts'
+import { StripToCodec } from './strip.ts'
 
 // ------------------------------------------------------------------
 // Assert
@@ -61,7 +62,11 @@ function Assert(context: TProperties, type: TSchema, value: unknown): unknown {
 // ------------------------------------------------------------------
 /** Executes Encode callbacks only */
 export function EncodeUnsafe(context: TProperties, type: TSchema, value: unknown): unknown {
-  const sorted = Settings.Get().unionPrioritySort ? UnionPrioritySort(type) : type
+  // Strip to codec-bearing paths first, then sort only what survives — the sort
+  // (UnionPrioritySort) rebuilds the whole schema, so running it on the pruned
+  // schema is cheaper.
+  const stripped = StripToCodec(context, type)
+  const sorted = Settings.Get().unionPrioritySort ? UnionPrioritySort(stripped) : stripped
   return FromType('Encode', context, sorted, value)
 }
 // ------------------------------------------------------------------
--- a/src/value/codec/strip.ts
+++ b/src/value/codec/strip.ts
@@ -1,0 +1,97 @@
+/*--------------------------------------------------------------------------
+
+TypeBox
+
+The MIT License (MIT)
+
+Copyright (c) 2017-2026 Haydn Paterson
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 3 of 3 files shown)


amit-evron-cyera:
Hey @sinclairzx81, this PR is a follow-up from [this comment](https://github.com/sinclairzx81/typebox/issues/1625#issuecomment-4739334969). Did you happen to take a look by any chance? 

Best regards

sinclairzx81:
Hi @amit-evron-cyera, sorry for the delay (I had been quite focused on the 1.3.0 release), but I wanted to get to this sooner. 

> Did you happen to take a look by any chance?

Yes, I have looked at the PR, but unfortunately I won't be able to merge this one in, at least not in this form. I will explain a bit why below, as this functionality has broader crossover with TypeBox's future design and optimization story moving towards 2.x, but won't be able to accept as of today.

## Reasoning

The main reason I'm hesitant to merge this one is mostly due to "implicit" schema pre-processing and transformation being an opaque operation the end user would have no control over, so if things go wrong, the user has no way to work around it. This also applies to the current `UnionPrioritySort`, which is configurable for now (to remain non-breaking), but will be removed in future versions of TypeBox.

Provisions have already been made to make priority sorting an API of `Type.*` (available in 1.3.0).

[Type.Priority Example](https://www.typescriptlang.org/play/?target=99&module=7#code/JYWwDg9gTgLgBAKjgQwM5wCoE8wFM4BmUEIcA5DDrgEYQAeZAUIwMYQB2q8GcAvJlQB0AVXbAOACgDa2PIIAKUcUsrS46gPQa4i3GGItcqVMHYBzOMHS46YADbAWweMnYATOBDAxx7ZHcZ1WVxBAGUWJW8JAAMAbzg6AC44dgBXEGpcKAAaOCxktIysuABfaIBKbMCBOXDImBj4pJT0zKhSiqqgoTrgKLiEgtas3PyWopy4AC8hiY7KxgBdcuXq9XWNza3tnfWtODZObmTgwQxRXwAeKRkhDAB5agArXBYYS9i13e+f7f31ZqnDAAOWGUAA3F9ftCfv88ic7qCJpCYajYdp1DMaiEQWCUWiCVt9iUAHy5IGPF5vD5QwkEuGAxF42l01FwsZApFtfGswnEsnYs6U17vT68ukMhFyXHIlni9GlEmLRYk5igSCwRAodAANX8qXwRBI5EoeFodA0ADd9bgmKwOFw4ABBPhwPV2A2CADCdlwrgkGFysRK5V++wAkuAHE54BcOIplM4sKFoPAAO7AOx2OCZOBQXAgCCW3BuOXy3b7I2kH1+9i5AAirwgblwuVcHgAouw2C3LOxCKkYKl82Xy39tMWoCYHUA)

```typescript
import * as Type from 'typebox'

const T = Type.Union([Type.Priority([   // Preprocessing is explicit and optional
  Type.Script(`{ x: number, y: number }`),
  Type.Script(`{ x: number }`),
  Type.Script(`{ x: number, y: number, z: number }`),
])])
                                        // const T: Type.TUnion<[[Type.TObject<{
                                        //   x: Type.TNumber;
                                        //   y: Type.TNumber;
                                        //   z: Type.TNumber;
                                        // }>, Type.TObject<{
                                        //   x: Type.TNumber;
                                        //   y: Type.TNumber;
                                        // }>, Type.TObject<{
                                        //   x: Type.TNumber;
                                        // }>]]>

import * as Value from 'typebox/value'

const A = Value.Clean(T, {})            // Implicit UnionPrioritySort will be removed
                                        // from Clean, Decode, and Encode in future
                                        // versions
```

Other optimizers will need to be moved outside of the Value.* operations and organized accordingly.

## StripToCodec

In terms of StripToCodec, I do think there is value in exploring optimizing operations like StripToCodec, but in terms of design, the function really needs to be an external call rather than an implicit operation of Encode/Decode. From an implementation standpoint, the following is the general thinking.

```typescript
// --------------------------------------------------------------------
// User Land (Passed to Framework)
// --------------------------------------------------------------------

const T = Type.Object({ .... }) // User-defined schema with embedded codec

// --------------------------------------------------------------------
// Framework Level (Customizable Optimizations)
// --------------------------------------------------------------------

import { Decode, Encode } from 'typebox/value'
import { StripToCodec } from './optimizers/strip-to-codec.ts'

export function FrameworkDecoder(type: TSchema)  {
  const stripped = StripToCodec({}, type)    // Explicit call before decode
  return (value: unknown) => Decode({}, stripped, value)
}
export function FrameworkEncoder(type: TSchema)  {
  const stripped = StripToCodec({}, type)    // Explicit call before encode
  return (value: unknown) => Encode({}, stripped, value)
}
```

The above works with the general direction of TypeBox; however, the problem with incorporating StripToCodec as a public API is that the function would need to exist as part of a larger suite of possible optimizers (and be managed as one of potentially many).

Accommodating a singular StripToCodec function really isn't possible today, as TypeBox has no concept of a schema optimizer. That said, it's absolutely something I would like to explore in the future, but it's not clear to me at this stage what optimizers could exist, and what parts of TypeBox would need to be changed in order to accommodate a broader range of them. This would take some design work and planning, and at this stage, there isn't much bandwidth available to explore the concept in depth.

## Considerations

Based on the implementation provided in this PR, it does seem like StripToCodec can be moved outside of Encode/Decode and called before these functions, so I think for now it's going to be better to hold off on the PR, but as mentioned, I am happy to start a discussion on the StripToCodec optimization itself.

I'll leave this PR open for now for follow-up discussions on the optimization, but will likely close it out in a few days. I'm very happy to field any follow up discussions on this one.

Cheers,
S

sinclairzx81:
@amit-evron-cyera Heya,

Thanks again for the PR, but will close this one out as not-planned (at least for now)

---

Would still be interested in a general broader discussion on schema optimizations for both Value.* operations and Compile.*, (which could include StripToCodec), but would need to establish the function as a larger set of possible functions. Figuring out what they could be would be an interesting topic (and I expect the scope could be quite broad)

Example: Optimizing Schematics to Reduce Validation Checks

```typescript
// ----------------------------
// unoptimized
//
// 3x typeof 'number' checks
// 3x range checks
//
// ----------------------------
const A = { 
  allOf: {
    { type: 'number', minimum: 0 }, // discarded (out of range)
    { type: 'number', minimum: 10 },
    { type: 'number', maximum: 100 },
  }
}
// ----------------------------
// optimized (50% reduced checks)
//
//  1x typeof 'number' check
//  2x range checks
//
// ----------------------------
const B = { type: 'number', minimum: 10, maximum: 100 }
```

Feel free to open a discussion if you would like to discuss StripToCodec in the context of other potential optimizations so they can be classified and considered for integration. Will consider them for 2.x

All the best!
S

PR #1652, opened 2026-07-23
Fix keyword guards for multi-type schemas

### LINKED ISSUE #1653: Schema.Compile evaluates type-specific keywords for null
`Schema.Compile` incorrectly evaluates array-specific keywords for a `null` instance when `type` includes both `array` and `null`.

TypeBox 1.3.6 reproduction:

```ts
import Schema from 'typebox/schema'

const schema = {
  type: ['array', 'null'],
  items: { type: 'string' }
}

Schema.Check(schema, null) // true
Schema.Compile(schema).Check(null) // throws: Cannot read properties of null (reading 'every')
```

The compiled expression effectively becomes:

```ts
(Array.isArray(value) || value === null) && value.every(...)
```

The same issue affects mixed `object | null`, `string | null`, and `number | null` schemas with their corresponding type-specific keywords. Those keywords should be guarded when the schema permits other instance types.

I have a tested fix ready and can submit it after discussion. I initially opened #1652 before noticing the contribution guidance; that PR is now closed.

orifrumer:
Fixes #1653

## Summary

Guard type-specific keyword checks when a schema's `type` array contains values outside that keyword's instance type.

## Problem

For this valid JSON Schema:

```json
{
  "type": ["array", "null"],
  "items": { "type": "string" }
}
```

`Schema.Check(schema, null)` returns `true`, but `Schema.Compile(schema).Check(null)` throws:

```text
TypeError: Cannot read properties of null (reading 'every')
```

The compiler currently treats the presence of `"array"` anywhere in the type array as sufficient to emit unguarded array keyword checks. It therefore generates logic equivalent to:

```js
(Array.isArray(value) || value === null) && value.every(...)
```

JSON Schema type-specific keywords only apply to compatible instances, so `items` must not be evaluated for `null`.

## Implementation

Only elide a type-specific runtime guard when every declared type is that exact type. Multi-type schemas retain the existing runtime guard, while single-type schemas keep the current fast path.

The same helper is used for object, array, string, and numeric keyword groups, preventing equivalent failures for other mixed-type schemas.

## Validation

- `deno task test` — 35,137 passed
- `deno task lint` — passed
- `deno fmt --check src/schema/engine/schema.ts test/typebox/runtime/schema/compile.ts` — passed
- `deno task build` — passed, including package type checks

### DIFF
--- a/src/schema/engine/schema.ts
+++ b/src/schema/engine/schema.ts
@@ -79,9 +79,9 @@ import { BuildUnevaluatedProperties, CheckUnevaluatedProperties, ErrorUnevaluate
 import { BuildUniqueItems, CheckUniqueItems, ErrorUniqueItems } from './uniqueItems.ts'
 
 // ----------------------------------------------------------------
-// HasOnlyTypeName
+// HasTypeName
 // ----------------------------------------------------------------
-function HasOnlyTypeName(schema: Schema.XSchemaObject, typename: string): boolean {
+function HasTypeName(schema: Schema.XSchemaObject, typename: string): boolean {
   return Schema.IsType(schema) &&
     (G.IsArray(schema.type) && schema.type.length > 0 && schema.type.every(type => G.IsEqual(type, typename)) ||
       G.IsEqual(schema.type, typename))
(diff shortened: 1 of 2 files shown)


orifrumer:
Closing because I missed the contribution guidance requiring an issue before a pull request. I will open a focused bug report and wait for maintainer discussion before submitting another PR.

sinclairzx81:
@orifrumer awesome, thank you! :)

sinclairzx81:
Hi @orifrumer, 

Thanks again for this PR, this really was a great catch, so really appreciate.

Updates have been published out on `1.3.7`. 

All the best!
S

orifrumer:
@sinclairzx81 
My pleasure :) and thanks for maintaining TypeBox as open source for the community. I appreciate the quick review and release!

PR #1655, opened 2026-07-23
Record contains-matched indices so unevaluatedItems accepts them

spokodev:
### Problem

Under JSON Schema draft 2020-12 §11.2, `unevaluatedItems` treats an item as
*evaluated* if it was matched by `prefixItems`, `items`, **or `contains`**.
TypeBox records evaluated indices for `prefixItems` and `items`
(`context.AddIndex(...)`), but `contains` used `value.some(...)` (interpreter)
and a compiled `.some(...)` and recorded nothing. So a `contains`-matched item
was treated as unevaluated, and a valid array was wrongly rejected — on both the
`Value.Check` and `Compile().Check` paths:

```ts
const S = { contains: { type: 'string' }, unevaluatedItems: false }
Value.Check(S, ['foo'])       // false — should be true
Compile(S).Check(['foo'])     // false — should be true
Value.Check({ prefixItems: [true], ...S }, [1, 'foo']) // false — should be true
```

`['foo']` has its only item matched by `contains`, so nothing is unevaluated and
`unevaluatedItems: false` is satisfied.

TypeBox already vendors this exact expectation in
`test/jsonschema/cases/v1/_unevaluatedItems.json` ("second item is evaluated by
contains" → `valid: true`), but that file is `_`-prefixed and excluded from the
active suite, so the behavior was never enforced. The official
JSON-Schema-Test-Suite `draft2020-12/unevaluatedItems.json` agrees.

### Fix

Record the index of every `contains`-matching item, in both `CheckContains` and
`BuildContains`. The compiled path uses `reduce` instead of `some` when
unevaluated tracking is active so the scan does not short-circuit before all
matching indices are recorded; when tracking is off it keeps the original
`.some(...)` fast-path. `minContains: 0` semantics are preserved (an
all-non-matching array with `minContains: 0` still validates).

### Test

Added a `Value.Check.Array` case (which exercises both the `Compile` and `Value`
paths via the shared `Ok`/`Fail` helper) asserting that a `contains`-matched item
counts as evaluated. Fails on `main`, passes with the fix; the full suite is
otherwise unchanged.

### DIFF
--- a/src/schema/engine/contains.ts
+++ b/src/schema/engine/contains.ts
@@ -49,16 +49,35 @@ export function BuildContains(stack: Stack, context: BuildContext, schema: Schem
   if (!IsValid(schema)) return E.Constant(true)
   const item = Unique()
   const isLength = E.Not(E.IsEqual(E.Member(value, 'length'), E.Constant(0)))
-  const isSome = E.Call(E.Member(value, 'some'), [E.ArrowFunction([item], BuildSchema(stack, context, schema.contains, item))])
+  if (!context.UseUnevaluated()) {
+    const isSome = E.Call(E.Member(value, 'some'), [E.ArrowFunction([item], BuildSchema(stack, context, schema.contains, item))])
+    return E.And(isLength, isSome)
+  }
+  // When unevaluated tracking is active, record the index of every matching
+  // item so unevaluatedItems treats them as evaluated. reduce (not some) is
+  // used so the scan does not short-circuit before all indices are recorded.
+  const index = Unique()
+  const accumulator = Unique()
+  const isSome = E.Call(E.Member(value, 'reduce'), [
+    E.ArrowFunction([accumulator, item, index], E.Ternary(BuildSchema(stack, context, schema.contains, item), context.AddIndex(index), accumulator)),
+    E.Constant(false),
+  ])
   return E.And(isLength, isSome)
 }
 // ------------------------------------------------------------------
 // Check
 // ------------------------------------------------------------------
 export function CheckContains(stack: Stack, context: CheckContext, schema: Schema.XContains, value: unknown[]): boolean {
-  if (!IsValid(schema)) return true
-  return !G.IsEqual(value.length, 0) &&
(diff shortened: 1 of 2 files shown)


sinclairzx81:
@spokodev Hi, thanks for another great PR!

The implementation here is good, but would need to apply a couple of additional updates to ensure the implementation is using symmetric guards (Guard and EmitGuard). I think what I'll do is I'll merge this one in as is, then apply symmetric guards over the top. I will include the additional Array test also. 

## Specification Sync

Btw, if you need to sync the latest specification, you can do so by running the following. TB will download and segment passing and failing tests (where failing tests get prefixed via `_`). You can use the `spec` command to pull the latest spec, and use `git diff` to see that's improved.

```bash
deno task spec # clones and segments passing and failing sets

deno task test  # run test against passing set
```

For compiler updates, running the `spec` command before the PR can help highlight which parts of the test suite are corrected by any updates. This is usually better than implementing specific compiler / value tests. Also, if the specification suite is missing a test, that's usually a sign TB needs to include a local test, but may also indicate that the upstream test suite itself could benefit from an additional test (those would get submitted [here](https://github.com/json-schema-org/JSON-Schema-Test-Suite))

## Specification Coverage

> TypeBox already vendors this exact expectation in test/jsonschema/cases/v1/_unevaluatedItems.json ("second item is evaluated by contains" → valid: true), but that file is _-prefixed and excluded from the active suite, so the behavior was never enforced. The official JSON-Schema-Test-Suite draft2020-12/unevaluatedItems.json agrees.

TB is pushing towards 100% coverage, and have been tracking coverage rates at the following....

https://github.com/sinclairzx81/json-schema-compliance-suite

Interestingly, the failing `contains` test was actually added a few months back, so TB needed to update for that one (this PR was most welcome!). Will be aiming for additional coverage hopefully before the end of the year, with `$dynamicRef` being one of the main outstanding. I certainly could use some help with that if you're interested :)

---

Will merge this one through, apply guard updates and publish out later today.
Thanks again!
S

sinclairzx81:
Hi @spokodev, Updates have been published on `1.3.8`.

Thanks again for your help!
S

Issue #1617, opened 2026-06-13
Add `sideEffects: false` to the `package.json`


adamscybot:
There was a chat about this [over here](https://github.com/sinclairzx81/typebox/issues/907) but in the context of the previous-gen package. It looks like this got lost during the transition.

Adding `sideEffects: false` (or whatever `sideEffects` declaration would be accurate - I don't think new-typebox has any sideeffects anywhere...?) would be useful where bundler tooling takes it into account (e.g. [bun bundler](https://bun.com/docs/bundler#optimizeimports), vite etc) and thus can minimize the typebox runtime that ends up in the final bundle to the minimum slice required for the consuming project, without any specific consumer-side build config/flag. 

---

By the way, I'm very close to releasing/making-public a typebox-related package: `bun-typebox-plugin`. It's a plugin for bun bundler (I will probably adapt to esbuild also) that is able to identify `Compile()` callsites from `typebox/schema` and inline the performant validator code at bundle-time, the reference to which is injected transparently into the validator, such that it removes the tradeoffs in eval-restricted environments -- at least where the schema/context inputs are statically-analyzeable (works across files as long as everything is esm) and safe.

There's a whole other bunch of features I am working on and potential ideas. Some of which could be enabled or optimized by exposing/enhancing some specific areas of the existing internal apis. It'd be great to discuss this with you/show what I'm up to -- if you have the time of course -- and let me know the preferred communication channel if so? Thanks!

hiSandog:
Before declaring the whole package side-effect-free, it would be worth checking every exported subpath for top-level registry mutation, polyfills, or initialization that consumers may rely on. A small bundler fixture can import one schema helper from each public entry point with Bun, Vite/Rollup, and esbuild, then verify both that unused modules disappear and that used runtime behavior still initializes correctly. That guards the package metadata itself, which unit tests running unbundled would not exercise.

sinclairzx81:
@hiSandog @adamscybot

Hi, sorry for the delay in response.

> Before declaring the whole package side-effect-free, it would be worth checking every exported subpath for top-level registry mutation, polyfills, or initialization that consumers may rely on. A small bundler fixture can import one schema helper from each public entry point with Bun, Vite/Rollup, and esbuild, then verify both that unused modules disappear and that used runtime behavior still initializes correctly. That guards the package metadata itself, which unit tests running unbundled would not exercise.

[...]

PR #1659, opened 2026-07-28
Expose Metrics counters as accessors

### LINKED ISSUE #1658: Schema construction fails when `Metrics` is frozen
## Problem

`Memory.Create` increments a counter on the module-level `Metrics` object:

```js
// build/system/memory/create.mjs:25
export function Create(hidden, enumerable, options = {}) {
  Metrics.create += 1
  ...
```

`Metrics.create` is a data property, so once `Metrics` is frozen that increment throws in strict mode. `Assign`, `Clone`, `Discard` and `Update` increment their own counters the same way, and every `Type.*` constructor routes through at least one of them — `Type.String()` via `Create`, `Type.Optional()` and `Type.Unsafe()` via `Clone` and `Update`. So this is not a case of metrics going stale; **all schema construction fails**:

```
TypeError: Cannot assign to read only property 'create' of object '#<Object>'
    at Create (typebox/build/system/memory/create.mjs:25:11)
    at String (typebox/build/type/types/string.mjs:14:11)
```

Because schemas are usually declared at module scope, this tends to surface as module initialisation throwing rather than as a failure at a call site.

## Why anything would freeze it

A host that shares one module graph between isolated consumers has to freeze it, otherwise one consumer can observe or corrupt another's module state through the shared instance. Deep-freezing the shared exports is the standard way to enforce that.

We hit this in `@temporalio/worker`. It runs each workflow in its own V8 sandbox, each with a private copy of the module graph. That isolation regressed with webpack >= 5.108.0 — sandboxes ended up sharing module state, so one workflow could observe another's — and v1.20.1 restored it, along with the per-workflow copy of the graph. For us that came to hundreds of megabytes and OOM-killed the worker. The SDK's escape hatch shares nominated modules across all sandboxes and deep-freezes them, so a sandbox cannot mutate state another can see — on a best-effort basis, since a freeze cannot reach closure variables, `Map`/`Set` contents or private fields. Sharing anything that reaches TypeBox then throws on the first schema construction.

The same transitive-freeze discipline appears in SES / Hardened JavaScript, where `harden()` freezes an object graph so untrusted code cannot tamper with objects it shares with its host.

## Reproduction

No framework needed. `deepFreeze` below is copied verbatim from `@temporalio/worker` — it is the code that broke for us, and it is what a deep freeze generally looks like.

```js
import * as System from 'typebox/system'
import Type from 'typebox'

const { Memory } = System

function deepFreeze(object, visited = new WeakSet()) {
  if (object == null || visited.has(object) || (typeof object !== 'object' && typeof object !== 'function')) return object
  visited.add(object)
  if (Object.isFrozen(object)) return object
  if (typeof object === 'object') {
    for (const name of [...Object.getOwnPropertyNames(object), ...Object.getOwnPropertySymbols(object)]) {
      const value = object[name]
      if (value && (typeof value === 'object' || typeof value === 'function')) {
        try { deepFreeze(value, visited) } catch { /* not everything can be frozen */ }
      }
    }
  }
  return Object.freeze(object)
}

for (const exported of Object.values(System)) {
  try { deepFreeze(exported) } catch { /* module namespace objects can't be frozen */ }
}

console.log('Metrics frozen?', Object.isFrozen(Memory.Metrics)) // true
Type.String() // TypeError: Cannot assign to read only property 'create'
```

On `typebox@1.3.8` / Node 24 that prints `Metrics frozen? true` and then throws. `Object.freeze(Memory.Metrics)` on its own is the one-line version, though nothing would freeze only that object in practice.

## Suggested fix

Hold the counters in module scope and expose them as accessors. Freezing an object does not disable its accessors, and the closure variables are not reachable by any freeze, so the increments keep working:

```ts
let create = 0
// ...

export const Metrics: TMetrics = {
  get create() {
    return create
  },
  set create(value: number) {
    create = value
  },
  // ...
}
```

`TMetrics` is unchanged. `Object.keys(Metrics)` and `{ ...Metrics }` produce identical output before and after, so the counters stay readable, enumerable and spreadable.

Worth noting for the shared-module case: nothing in TypeBox reads the counters — they are five `+= 1` writes and a declaration, and no behaviour branches on them. So a counter shared between sandboxes has no observable effect on schema construction, even though a deep freeze cannot reach it.

I have this on a branch with a test that freezes `Metrics` and asserts the counters still increment. `deno task lint`, `deno task test` (35,188 passing) and `deno task build` all pass, and running the reproduction above against a pack of that build prints `Metrics frozen? true` followed by no error — the fix survives the freeze rather than avoiding it. Happy to open a PR if you would like one; the readme asks for an issue first, hence this.

## Alternative

A setting such as `Settings.Set({ metrics: false })` would also avoid the write, and would sit naturally alongside `useAcceleration`. The drawback is that it has to be set before the first schema is constructed, which a library depending on TypeBox cannot guarantee and code inside a sandbox may have no opportunity to do at all. The accessor change needs no configuration, which is why I have suggested it first — but I am happy to implement whichever you prefer.

---
🤖 Drafted with [Claude Code](https://claude.com/claude-code)

TastyPi:
Closes https://github.com/sinclairzx81/typebox/issues/1658

`Metrics` increments a counter for every `Assign`, `Create`, `Clone`, `Discard` and `Update` call. The counters are data properties, so the increment throws `TypeError` once `Metrics` is frozen — and since every `Type.*` constructor routes through at least one of those five, schema construction fails outright rather than merely losing its metrics. The issue has a self-contained reproduction.

This keeps the counters as module-scope variables and exposes them through accessors. A freeze cannot disable an accessor, so the increments keep working.

`TMetrics` is unchanged, and `Object.keys(Metrics)` and `{ ...Metrics }` produce identical output before and after, so the counters stay readable, enumerable and spreadable.

## Tests

Two tests in `test/typebox/runtime/system/memory.ts`: that the counters increment, and that they still increment after `Object.freeze(Memory.Metrics)`.

They assert the behaviour rather than the mechanism, so they do not pin the implementation to accessors specifically.

## Verification

- `deno task lint` — clean, 685 files
- `deno task test` — 35,188 passing, 0 failing
- `deno task build` — clean
- packed that build and ran the issue's reproduction against it: prints `Metrics frozen? true` and then does not throw, where `typebox@1.3.8` throws at `create.mjs:25`

## Note

Happy to drop this in favour of a `Settings` flag (`Settings.Set({ metrics: false })`) if you would rather metrics were opt-out than freeze-tolerant — the tradeoff is that a flag has to be set before the first schema is constructed, which a library depending on TypeBox cannot guarantee.

---
🤖 Drafted with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/system/memory/metrics.ts
+++ b/src/system/memory/metrics.ts
@@ -39,11 +39,48 @@ export interface TMetrics {
   update: number
 }
 
+// ------------------------------------------------------------------
+// Counters
+// ------------------------------------------------------------------
+// Metrics exposes these counters as accessors so they remain writable when
+// Metrics is frozen. Hosts that share one module graph between isolated
+// consumers deep-freeze it, and incrementing a frozen data property throws.
+let assign = 0
+let create = 0
+let clone = 0
+let discard = 0
+let update = 0
+
 /** TypeBox instantiation metrics */
 export const Metrics: TMetrics = {
-  assign: 0,
-  create: 0,
-  clone: 0,
-  discard: 0,
-  update: 0
+  get assign() {
+    return assign
+  },
+  set assign(value: number) {
+    assign = value
+  },
+  get create() {
+    return create
+  },
+  set create(value: number) {
+    create = value
+  },
+  get clone() {
+    return clone
+  },
+  set clone(value: number) {
+    clone = value
+  },
+  get discard() {
+    return discard
+  },
+  set discard(value: number) {
+    discard = value
+  },
+  get update() {
+    return update
+  },
+  set update(value: number) {
+    update = value
+  }
 }
--- a/test/typebox/runtime/system/memory.ts
+++ b/test/typebox/runtime/system/memory.ts
@@ -121,3 +121,22 @@ Test('Should Create 11', () => {
   Assert.PropertyIsEnumerable(B, '~unsafe')
   Settings.Reset()
 })
+// ------------------------------------------------------------------
+// Metrics
+// ------------------------------------------------------------------
+Test('Should Increment Metrics', () => {
+  const A = Memory.Metrics.create
+  Type.String()
+  Assert.IsEqual(Memory.Metrics.create, A + 1)
+})
+// ------------------------------------------------------------------
+// Metrics Should Increment When Frozen
+//
+// https://github.com/sinclairzx81/typebox/issues/1658
+// ------------------------------------------------------------------
+Test('Should Increment Metrics When Frozen', () => {
+  Object.freeze(Memory.Metrics)
+  const A = Memory.Metrics.create
+  Type.String()
+  Assert.IsEqual(Memory.Metrics.create, A + 1)
+})


sinclairzx81:
@TastyPi Thanks for the PR (and again apologies for the delay)

Will close out as per comment https://github.com/sinclairzx81/typebox/issues/1658#issuecomment-5613298121 but happy to discuss more based on the rationale documented on the issue thread. Basically, I need to find more cases of applications freezing imports before considering a change to Metrics, as the "freeze on import" sandbox strategy would be non-typical.

> There may be some cross over here with https://github.com/sinclairzx81/typebox/issues/1617 regarding whether TypeBox is side-effect free (and whether or not Metrics constitutes side-effects); the import freeze and bundle side effects seem somewhat related.

Will close out for now, but again happy to discuss more on the issue thread.
All the best!
S

PR #1672, opened 2026-08-23
Fix (Memory): Ensure Immutability for Assign, Discard and Update

webstrand:
When `immutableTypes` is true, `Memory.Create` freezes every schema it produces, but schema produced by `Memory.Update` are not frozen.

```ts
import Type from 'typebox'
import { Settings } from 'typebox/system'

Settings.Set({ immutableTypes: true })

const N = Type.Number()
const O = Type.Optional(Type.Number())

Object.isFrozen(N)  // true
Object.isFrozen(O)  // false
```

### DIFF
--- a/src/system/memory/assign.ts
+++ b/src/system/memory/assign.ts
@@ -30,10 +30,11 @@ THE SOFTWARE.
 // deno-fmt-ignore-file
 
 import { Metrics } from './metrics.ts'
+import { Freeze } from './freeze.ts'
 
 type ObjectLike = Record<PropertyKey, any>
 
-/** 
+/**
  * Performs an Object assign using the Left and Right object types. We track this operation as it
  * creates a new GC handle per assignment.
  */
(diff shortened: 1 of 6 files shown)


sinclairzx81:
[@webstrand](https://github.com/webstrand) Hi! Thanks for the PR, this was a great catch! :)

I've made a small update (added a new `freeze.ts` module) that centralizes the freeze logic across Create, Update, Assign and Discard (where Discard and Assign are Update-like operations that return a new objects), so they should use conditional freeze too. Everything else was cool.

---

### Memory.Clone

I did actually consider Clone as well, but have left that function as is for now. I note the Clone function isn't actually called directly from `Type.*`, but is currently serving as the backend logic used by `Value.Clone()`, so it can't reasonably freeze values there.

> Note: I do think there may be potential for Clone to retain the frozen state of the input value (by checking `Object.isFrozen(...)` on the input), but I'll defer that for another time. There are some considerations around performance, and whether TypeBox should diverge too much from `structuredClone(...)`. Updates to Clone need a bit more thinking overall, but it's very tempting to just move to `structuredClone(...)` in V2 to simplify.
>
> Will review Clone in later revisions.

---

Will merge this one in now, and will publish out later tonight along with some additional spec updates; and again, thanks for the PR! Finding and helping to fix issues like this is super appreciated! :)

All the best!
S

sinclairzx81:
Hi @webstrand, Updates have been published on 1.3.18. Thanks again! :)

Issue #1519, opened 2026-01-23
t.Array accept sparse array and does not validate the schema


Teyik0:
I am mostly using Elysia but seems like the issue comes from typebox.
Typebox Check send true while only filling one element of the array. I was expecting a validation error on this.

Here is the repro.
```ts
import { describe, it, expect } from "bun:test";
import { t } from "elysia";
import { Value } from "@sinclair/typebox/value";

describe("TypeBox Sparse Array Validation Bug", () => {
  const schema = t.Array(t.Object({ name: t.String() }));

  it("should fail validation for sparse arrays (but doesn't - BUG)", () => {
    // Create a sparse array with empty slots at indices 0-8
    const sparseArray: { name: string }[] = [];
    sparseArray[9] = { name: "ninth" };

    const checkResult = Value.Check(schema, sparseArray);
    const errors = [...Value.Errors(schema, sparseArray)];

    // EXPECTED: Check() should return false because there are validation errors
    expect(checkResult).toBe(false); // This will FAIL - demonstrates the bug
    expect(errors.length).toBe(9); // This passes - errors are correctly detected
    expect(sparseArray.length).toBe(10);
  });

  it("should fail validation for arrays with explicit nulls (works correctly)", () => {
    const withNulls = [null, null, { name: "third" }];

    const checkResult = Value.Check(schema, withNulls);
    const errors = [...Value.Errors(schema, withNulls)];

    // This works correctly
    expect(checkResult).toBe(false); // ✅ Correct!
    expect(errors.length).toBe(2);
  });

  it("shows Value.Errors() detects all empty slots correctly", () => {
    const sparseArray: { name: string }[] = [];
    sparseArray[9] = { name: "ninth" };

    const errors = [...Value.Errors(schema, sparseArray)];

    // Should have 9 errors (one for each empty slot at indices 0-8)
    expect(errors.length).toBe(9);

    // Verify each error corresponds to an empty slot
    errors.forEach((err, i) => {
      expect(err.path).toBe(`/${i}`);
      expect(err.message).toBe("Expected object");
    });
  });

  it("demonstrates inconsistency: Check() should match Errors()", () => {
    const sparseArray: { name: string }[] = [];
    sparseArray[5] = { name: "fifth" };

    const checkResult = Value.Check(schema, sparseArray);
    const hasErrors = [...Value.Errors(schema, sparseArray)].length > 0;

    // EXPECTED: Check() and Errors() should agree
[...]

PR #1675, opened 2026-08-26
Fix (Schema): Compiled validator skips array holes

dylanpulver:
Fuzzed the compiled validator against `Schema.Check` over 3000 random array schemas per seed. Six seeds disagreed 40, 59, 62, 67, 53 and 69 times. Every disagreement involved an array with holes. Removing holes from the generator drops the same run to 0, which isolates the cause.

`EmitGuard.Every` emitted `value.every(...)` when the offset was 0, and `EmitGuard.Some` emitted `value.some(...)`. Both native methods skip holes. `Guard.Every` and `Guard.Some` iterate by index and read a hole as `undefined`, so the compiled validator saw fewer elements under `items` and `unevaluatedItems`. `contains` diverged in the opposite direction.

```ts
const T = Type.Array(Type.Number())
Value.Check(T, new Array(3))      // false
Compile(T).Check(new Array(3))    // true
```

`Type.Array(Type.Never())` also accepted `new Array(3)` once accelerated, and `Compile(T).Errors(value)` returned nothing where `Value.Errors` reported 8 errors. This is the shape from #1519 on the compiled path.

`Every` emits an index loop when the offset is non-zero, so this makes the offset-0 path match the branch beside it.

After the change the same six seeds report 0. Full suite green at 38506. Reverting `emit.ts` alone fails 7 of the 8 added tests.

Two divergences remain that this diff does not touch, both reproducing on a clean checkout:

- `uniqueItems` disagrees between `Check` and `Errors` on holes
- `required: ['toString']` alongside `properties` and `additionalProperties: false`


sinclairzx81:
@dylanpulver Hi! thanks for the PR!

The varying results for Compile and Check is a bug and needs to be fixed, but I am wondering how best to fix it. TBV1 doesn't have a strong specification for handling sparse arrays, so would be a good time establish one.

> Note: Have made a small update to move the tests to the `Value.Array` suite as these currently define TypeBox level assertion behaviors (above JSON Schema). This suite will assert parity for Compile, Check and Errors.

---

### Enumeration of Empty Elements

My thinking is that Every and Some should correspond to the ECMA262 interpretation of `.every(...)` and `.some(...)`, and where these functions skip `<empty>` elements on enumeration. This skipping behavior for sparse `<empty>` elements seems to be true for all the `Array` functions.

```typescript
const input:  unknown[]  = new Array(5)
const output: unknown[] = []

input[3] = 42

input.forEach(input => output.push(input))   // <empty> elements are skipped here, this also
                                             // applies to .some() and .every().

console.log(output)                          // [42]
```

---

### Ecma262 Algorithm

This is the algorithm for `.every()` where the specification does skip on condition `c. If kPresent is true, then`.

https://tc39.es/ecma262/#sec-array.prototype.every


1. Let obj be ? [ToObject](https://tc39.es/ecma262/#sec-toobject)(this value).
2. Let length be ? [LengthOfArrayLike](https://tc39.es/ecma262/#sec-lengthofarraylike)(obj).
3. If [IsCallable](https://tc39.es/ecma262/#sec-iscallable)(callback) is false, throw a TypeError exception.
4. Let k be 0.
5. Repeat, while k < length,
   a. Let propertyKey be ! [ToString](https://tc39.es/ecma262/#sec-tostring)([𝔽](https://tc39.es/ecma262/#%F0%9D%94%BD)(k)).
   b. Let kPresent be ? [HasProperty](https://tc39.es/ecma262/#sec-hasproperty)(obj, propertyKey).
   c. If kPresent is true, then
      i. Let kValue be ? [Get](https://tc39.es/ecma262/#sec-get-o-p)(obj, propertyKey).
      ii. Let testResult be [ToBoolean](https://tc39.es/ecma262/#sec-toboolean)(? [Call](https://tc39.es/ecma262/#sec-call)(callback, thisArg, « kValue, [𝔽](https://tc39.es/ecma262/#%F0%9D%94%BD)(k), obj »)).
      iii. If testResult is false, return false.
   d. Set k to k + 1.
6. Return true.

---

### Revised Implementation

Based on the above algorithm, rather the removing the `.every()` / `.some()` calls, it may be better to:

1) retain the `.every()` assertion for when `offset` is `0`. (standard ECMA262 behavior)
2) update the `for-loop` to skip via `if(!(index in array)) continue` check per element when `offset` is `> 0`

This with the goal of mirroring the ECMA 262 behaviors ...

---

What are your thoughts?

sinclairzx81:
Hi @dylanpulver, just a quick follow up here.

I was taking a look at the `main` branch and testing a couple of implementations. From what I can tell, the following would be the updated guards to better match the ECMA262 sparse array behavior (adding in the `index in value` guard), however the implementation would require the implemented test assertions to be updated (as the implementation no longer interprets `<empty>` as `undefined`)

### Guard

```typescript
// --------------------------------------------------------------------------
// Array
// --------------------------------------------------------------------------
/** Returns true if every element from offset satisfies the callback, short-circuiting on the first failure */
export function Every<T>(value: T[], offset: number, callback: (value: T, index: number) => boolean): boolean {
  for (let index = offset; index < value.length; index++) {
    if ((index in value) && !callback(value[index], index)) return false
  }
  return true
}
/** Returns true if every element from offset satisfies the callback, using exhaustive enumeration */
export function EveryAll<T>(value: T[], offset: number, callback: (value: T, index: number) => boolean): boolean {
  let result = true
  for (let index = offset; index < value.length; index++) {
    if ((index in value) && !callback(value[index], index)) result = false
  }
  return result
}
/** Returns true if some element satisfies the callback, short-circuiting on the first success */
export function Some<T>(value: T[], callback: (value: T, index: number) => boolean): boolean {
  // note: could use value.some(callback) instead of for-loop here
  for (let index = 0; index < value.length; index++) {
    if ((index in value) && callback(value[index], index)) return true
  }
  return false
}
/** Returns true if some element satisfies the callback, using exhaustive enumeration */
export function SomeAll<T>(value: T[], callback: (value: T, index: number) => boolean): boolean {
  let result = false
  for (let index = 0; index < value.length; index++) {
    if ((index in value) && callback(value[index], index)) result = true
  }
  return result
}
```

### EmitGuard

```typescript
// --------------------------------------------------------------------------
// Array
// --------------------------------------------------------------------------
export function Every(value: string, offset: string, params: [value: string, index: string], expression: string): string {
  return G.IsEqual(offset, '0')
    ? `${value}.every((${params[0]}, ${params[1]}) => ${expression})`
    : `((value, callback) => { for(let index = ${offset}; index < value.length; index++) if ((index in value) && !callback(value[index], index)) return false; return true })(${value}, (${params[0]}, ${params[1]}) => ${expression})`
}
export function Some(value: string, params: [value: string, index: string], expression: string): string {
  return `${value}.some((${params[0]}, ${params[1]}) => ${expression})`
}
export function SomeAll(value: string, params: [value: string, index: string], expression: string): string {
  return `((value, callback) => { let result = false; for(let index = 0; index < value.length; index++) if ((index in value) && callback(value[index], index)) result = true; return result })(${value}, (${params[0]}, ${params[1]}) => ${expression})`
}
export function Counted(value: string, params: [value: string, index: string], expression: string): string {
  return `${value}.reduce((result, ${params[0]}, ${params[1]}) => (${expression}) ? ++result : result, 0)`
}
```

---

TypeBox needs to settle on one implementation. I tend to lean towards the above ECMA262 interpretation as that's generally the expected behavior for `.every()` and `.some()`. It also has the better performance as JavaScript runtimes have a better time optimizing through non `<empty>` iteration on arrays. I did note the performance drops to around 60% using the `for-loop-only` implementation as the runtime needs to visit potentially `<empty>` elements, so need to be mindful of that too.

---

Again, let me know your thoughts on any of the above.
Cheers
S

sinclairzx81:
@dylanpulver Hi, thanks again for this PR

I've decided to close this one out out as I have set up a separate PR where I've opted for the `.every(...)` and `.some(...)` enumeration (in line with the above-mentioned ECMA262 logic), and have created some new test specifications in line with that view of things.

The PR mostly addresses the more critical issue of Compile and Check returning mismatched results (that bug needed to be fixed with some urgency) ... while settling on a de facto ECMA262 enumeration strategy which does not consider `<empty>` elements as real values.

Related PR: https://github.com/sinclairzx81/typebox/pull/1676

---

### Considerations

For what it's worth, I actually quite liked your PR!!! 😄

Unfortunately, I couldn't bring on the `for-loop` logic that was considering `<empty>` elements as `undefined`, as that behavior was at odds with the JavaScript `every(...)` function (and where the `Every(...)` guard is meant to represent `every(...)`), so this needed to be kept in sync.

There is also this aspect as it relates to JSON Schema

> The JSON Schema specification doesn't really have any notion of "arrays with holes", so any logic TypeBox implements here would be above and beyond the JSON Schema semantics. Conversely, implementing the `for-loop` logic would be beyond the ECMA262 specification for Array.prototype.every. Both aspects put TypeBox in a difficult position as there is no rule book for handling sparse array validation under either specification.

#### Element Item Counting

Opting for the ECMA262 implementation was the safe bet (because I can point to that specification), but to be honest, there were aspects of the `for-loop` implementation I liked, specifically the idea of moving to a `counted` non-empty element algorithm, and applying counts to array constraints such as `minItems`, `maxItems`, and `contains`.

Element counting is actually done for `minLength` and `maxLength` constraints (Unicode), which count UTF-16 / grapheme clusters, and as such counted assertions for non-empty sparse arrays could fit under a similar assertion model; however, given how rare sparse arrays are and the additional complexity required to count ... keeping the codebase as is while fixing the Check and Compiler disparity was considered the best call (at least for now).

---

Hope this brings some insight into things. I do need to fix the Compile / Check disparity pretty quickly, but the logic implemented for this PR raises some interesting other aspects which would be worth exploring in future.

Again, thanks for finding this issue and the PR, it was appreciated.
All the best
S

